PR #26, opened 2015-03-05
bug fix

tianxiaoliang:
1.check password
2.ommit white space in user name and password, otherwise there is
Exception "username/password is invalid"


majst01:
Hi,

thanks for your contribution, but from my opinion a API should not modify the given Input especially not username and password. It might be a right assumption in most cases, but in general a username and a password can contain trailing/leading white spaces. 

You should check the username and password in the caller.

Same applies for the url.

So i will not merge you pull request, sorry.

PR #27, opened 2015-03-12
Add iterator to Serie to iterate all point rows with no memory overhead

ivansenic:
Hi,

I noticed that the getRows() method that returns the list with map for each row is quite non-effective regarding memory footprint.. For each row we would have one map created, which would mean that for a series having 2k rows we would have 2k maps + the list.. That's really a lot of garbage considering that the .newHashMap() method in guava creates an empty hash map with default capacity of 16.. Considering that in most cases serie has only two points (time & value), that's just waist of space.. Especially if we are loading series having more rows (think about 50k).. Also time you need to construct a complete list of rows is quadratic (#numRows * #numColumns), which can be high when having big number of rows and columns..

So I proposed the simple solution of providing the iterator on the Row class.. All operations here are constant, except getPoint(String columnName) which has additional time of C/2 to find corresponding column index.. Regarding memory we are only creating one Row object per row, which has same size as Integer object (16 bytes)..

If you consider series with 2k rows and 2 columns, I am only creating 32k garbage with iterator, while with using getRows it's approximately 352k only for Maps, plus some 8k for list in addition..

### DIFF
--- a/src/main/java/org/influxdb/dto/Serie.java
+++ b/src/main/java/org/influxdb/dto/Serie.java
@@ -1,8 +1,10 @@
 package org.influxdb.dto;
 
 import java.util.Arrays;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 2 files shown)


ivansenic:
Additionally I can also provide the column iterator which would work in same way, just on the columns.. This might also be useful in some situations when you want to process column by column.. If you think it's a good idea, I can easily add it to the commit...

majst01:
Hi,

agreed but can you please omit the usage of org.apache.commons.lang.ArrayUtils, i dont want to introduce new dependencies.

Except that it looks ok.

Greetings
Stefan

ivansenic:
OK pushed with no ArrayUtils dependency.. I somehow assumed this was already a dependency, sorry.. I also changed the name of the method to get the iterator to rows(), seams more logical..

If anything else is needed let me know..

majst01:
Perfect, one more thing sorry, from my opinion bot Inner Classes RowIterator and Row should be static, right ?

If you agree, and update your pull request i will merge it

ivansenic:
Hi Stefan,

Both classes can not be static because they access the points field of the in-closing Serie instance.. 

I could make them static and pass the Serie object in the constructor of both classes, but I don't see what it's gained with it? What do you think?

majst01:
Hi,

is normally prefer static inner classes but im fine with your patch and merge it.
Thank you for your contribution !

PR #31, opened 2015-05-12
Updated dockerfile. Set influxdb package to v0.8.8

enric-sinh:
It was pulling the latest package, but the driver is not compatible yet with the latest 0.9, it only works up to 0.8.8.


majst01:
This for this, merging it

PR #42, opened 2015-06-27
Slight refactor on BatchProcess & fixed comments

jhorwit2:


majst01:
Hi,

got your point, but can you please give some numbers on how this patch increases performance ?

Greetings
Stefan

jhorwit2:
Hey,

This won't necessarily directly correlate to an increase in performance; however, it is cleaner IMO than creating synchronized lists & blocks of code. It's always best to let java libraries handle that when possible (like what blocking queues do)

- Josh

PR #50, opened 2015-07-03
Add public accessor methods to Point

ajsquared:
When writing tests for projects that use influxdb-java I've found myself wanting to be able to access the fields on a `Point` object.  This adds public accessor methods to allow doing so.

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -151,7 +151,7 @@ public class Point {
 	/**
 	 * @return the tags
 	 */
-	Map<String, String> getTags() {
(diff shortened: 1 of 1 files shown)


majst01:
Hi @ajsquared 

Thanks for interest and contribution to influxdb-java.
Adding public methods for testing only is a weak argument, i prefer narrow APIs with the least necessary visibility to keep it side effect free as possible.

For testing the resulting lineprotocol is visible which is the only purpose of a point object. You can either test againts this, or use reflection to access the properties in you tests.

Greetings 
Stefan

ajsquared:
Okay, that's fair enough.  Thanks!

PR #66, opened 2015-07-30
Set a point's defaut time as system time.

caoli5288:
Defaut time as 0 is a bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ BatchPoints batchPoints = BatchPoints
 				.build();
 Point point1 = Point.measurement("cpu")
 					.time(System.currentTimeMillis(), TimeUnit.MILLISECONDS)
-					.field("idle", 90L).field("user", 9L)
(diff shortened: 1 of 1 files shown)


src/main/java/org/influxdb/dto/Point.java
```diff
@@ -123,13 +123,12 @@ public Builder fields(final Map<String, Object> fieldsToAdd) {
...
 		 * @return the Builder instance.
 		 */
-		public Builder time(final long timeToSet, final TimeUnit precisionToSet) {
-			this.time = timeToSet;
-			if (null != precisionToSet) {
-				this.precision = precisionToSet;
-			}
-			return this;
-		}
+        public Builder time(final long timeToSet, final TimeUnit precisionToSet) {
```
majst01: Do not format the code in a different manner

majst01:
Can you please first create an issue which describes the problem and the reference the pull request to the ticket you address with it.

PR #67, opened 2015-07-30
Update README.md example

gunnaraasen:
Due to issue https://github.com/influxdb/influxdb/issues/3467, the field key named `user` makes InfluxDB return an error. This change prevents users from running into the bug, which will likely be fixed in 0.9.3.


majst01:
Good catch, fixed that already in the Tests, but forgot to adjust the Readme, thanks

PR #79, opened 2015-08-21
allow user to provide ThreadFactory for BatchProcessor

marsishandsome:
I'm implementing a JVM profiling tool, and I use InfluxDB to store the profiling data.
In my application, I do not want InfluxDB Client to allocate it's own thread.
So an addition argument (ThreadFactory) is added.

### DIFF
--- a/src/main/java/org/influxdb/impl/BatchProcessor.java
+++ b/src/main/java/org/influxdb/impl/BatchProcessor.java
@@ -3,12 +3,9 @@ package org.influxdb.impl;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 
+import com.google.common.util.concurrent.MoreExecutors;
 import org.influxdb.InfluxDB;
 import org.influxdb.dto.BatchPoints;
 import org.influxdb.dto.Point;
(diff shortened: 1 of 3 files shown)


marsishandsome:
@majst01 would you please review the patch?

src/main/java/org/influxdb/impl/BatchProcessor.java
```diff
@@ -3,12 +3,9 @@
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
majst01: I only want explicit imports, no wildcards please

src/main/java/org/influxdb/impl/BatchProcessor.java
```diff
@@ -126,13 +137,16 @@ public static Builder builder(final InfluxDB influxDB) {
...
 	BatchProcessor(final InfluxDBImpl influxDB, final int actions, final TimeUnit flushIntervalUnit,
-			final int flushInterval) {
+			final int flushInterval, final ThreadFactory threadFactory) {
 		super();
 		this.influxDB = influxDB;
 		this.actions = actions;
 		this.flushIntervalUnit = flushIntervalUnit;
 		this.flushInterval = flushInterval;
 
+		ScheduledThreadPoolExecutor service= (ScheduledThreadPoolExecutor) Executors.newScheduledThreadPool(1, threadFactory);
```
majst01: Is this cast necessary ?

marsishandsome: Yes, I thinks it's necessary.

majst01:
Can you please provide a Test which shows that current and new behavior is the same.

PR #87, opened 2015-09-15
Refactorings and test fixes

Kindrat:
Code refactorings:
- removed unnecessary 'this'
- removed unnecessary 'super'
- removed unnecessary 'super()' in constructors
- removed unnecessary 'final'
- recreated equals() and hashcode() in Query class using java7 style
- added slf4j loggers
- added error reason and error code to message in error handler
- proper exception handling in error handler - we need to know if something happened during response parsing
- extracted before and after test methods to abstract test class
- added logback test configuration - to see new logs in tests

Fixes:
- time builder call in InfluxDBTest.testWrite
- tags parsing in InfluxDBTest.testWrite - due to 0.9.3v changes we need to pass them in GROUP BY.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,19 @@
 		</pluginManagement>
 	</build>
 	<dependencies>
-		<dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jul-to-slf4j</artifactId>
+            <version>1.7.12</version>
+        </dependency>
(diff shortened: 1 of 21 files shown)


majst01:
I only accept pull requests which clearly fix one and only one problem which is described in a issue.
This pull request is a bunch of modifications which does a lot of different things. 

I really appreciate your work, but this is the wrong way to submit your work.

PR #90, opened 2015-09-16
Slf4j logging

Kindrat:
Replaced sout calls with slf4j logging, added logback configuration to test resources to show working logs in tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,16 @@
 		</pluginManagement>
 	</build>
 	<dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.12</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jul-to-slf4j</artifactId>
+            <version>1.7.12</version>
+        </dependency>
 		<dependency>
 			<groupId>org.testng</groupId>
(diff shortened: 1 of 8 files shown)


Kindrat:
Any comments?

majst01:
Hi,

You change code all over the place, reorganized imports, and introduced an additional dependency to influxdb-java just to get rid of one single line of sysout in production code.

As i said in the other pr where you transformed a statement to a java 8 lamba, this is not worth the hassle. Users of influxdb-java, and this applies for other libraries as well, are not willing to get dependencies forced because they use a library.

Im willing to accept pr´s which tries to solve the one sysout statement in the production code 

https://github.com/influxdb/influxdb-java/blob/master/src/main/java/org/influxdb/impl/InfluxDBImpl.java#L109

but this must be a oneliner or something like that.
A PR should be small an non disruptive, solve only one problem, does not change formatting nor imports, is easy to understand and closes an existing issue.

Greetings
Stefan

Kindrat:
Well, that is not "fixing one sout". In our project we use your lib and need logging. Sout is not logging, it's just throwing data away. Slf4j is logging facade, that allows devs to use their own log implementation (http://www.slf4j.org/). That is the simplest and most convenient way. I've tried to make minimal changes to existing code, so changed only sout calls and want to discuss adding more debug|info messages.

If the main problem is import rearrangement - sorry, that Intellij default behavior. I may revert and make another pr with log changes only.

PR #102, opened 2015-10-14
Escape equals signs in tag keys/values

agentgonzo:
If you put an equals sign in the Tag key/value, it would throw an 'unable to parse' RuntimeException when trying to write to influx


majst01:
Hi @agentgonzo 

nice catch, can you please add a Test, then im fine with this

agentgonzo:
OK, I've updated the commit with some tests. Let me know if there's anything else you need

PR #103, opened 2015-10-20
Allow buffering & data retention in BatchProcessor

andrewdodd:
These commits restructure the `BatchProcessor` to introduce some options for the buffering that (I hope) make it a little more useful.

I found that I was writing pretty similar wrappers around the unbuffered writer to get these options, but that they would probably be pretty common use cases.

The general changes are:
 - Allow configuration of queue capacity (either 'unbounded' or N)
 - If limited capacity (or runtime limit of unbounded), provide option for behaviour on writing when capacity is reached (i.e. throw exception, drop oldest record, or drop the current record)
 - Allow configuration of discard / retain on failed write of buffered points
 - If data retention is enabled, rate throttle calls to `write()` IF previous attempts to `write()` failed


majst01:
Hi,

Way to complex modifications. formatting changes all over the place, API change.

As i already mentioned, a PR should solve one single problem which is described in a issue.

I think some of the ideas are really useful but they need splitted into small understandably modifications, then im happy to merge them.

PR #104, opened 2015-10-20
Changing client insertion

andrewdodd:
This is a minor change to how the 'client' object is built.

This makes changing this file for Java 1.6 support more obvious.


majst01:
This seems to be a useful clarification, thanks.

PR #105, opened 2015-10-20
Java6 support

andrewdodd:
Changes for Java 6.

Please have a look, I'm aware that Java 6 support is probably a bit backward, but I need it so I thought I'd make the commits (since it seems like Java 6 support is not too hard for the 2.0 release) and you might keep it as a branch?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,8 +27,10 @@
 			<distribution>repo</distribution>
 		</license>
 	</licenses>
+	
 	<properties>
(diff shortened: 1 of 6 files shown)


majst01:
Hi,

this PR does not only change the codebase to support Java 6, it also changes the Exception Handling and the way retrofit creates a client connection. 

I dont plan to support Java 6 because this is EOL anyway, and i dont see why someone needs to stay with it especially in new projects.

PR in general should solve one single problem which was described in an issue. They also should not modify formatting of surrounding code.

So i will close this PR.

PR #108, opened 2015-10-21
Data retention enhancement

andrewdodd:
These are enhancements to allow data retention in the batch processor. Please see the commit for more info.

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -11,7 +11,7 @@ import org.influxdb.dto.Point;
 import org.influxdb.dto.Pong;
 import org.influxdb.dto.Query;
 import org.influxdb.dto.QueryResult;
-import org.influxdb.impl.BatchProcessor.BatchEntry;
+import org.influxdb.impl.BatchProcessor.BufferFailBehaviour;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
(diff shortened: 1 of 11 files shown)


Kindrat:
Consider making fork)

majst01:
I like your approach, but i need some time during the weekend to review it carefully.
Do you think it is worth to pull it now or create make a 2.1 version instead ?

andrewdodd:
Hey,

First, I think it is up to you. You are the primary maintainer, so in most ways it is your choice. I don't think I have enough of a 'different direction' to really warrant making a fork. If people really like my additions they can go to andrewdodd/influxdb-java, right?

I have been thinking about this set of commits, especially with respect to the existing library. In some ways, my additions are expanding the scope of the API quite a lot, and designing for the general case they are addressing will be hard. (i.e. at least the existing library 'forces' its users to think about data retention explicitly). However, I have to admit, I'm not a big fan of the 'auto-batching' option in the current library, precisely because it silently drops the data it is claiming to buffer and batch.

Regarding these commits, I had a few goes at getting to the design I wanted (as you can see from the commit log), so please look at the end state more than the intervening states. (I was trying to get too cute by picking the right underlying queue etc, but in the end I just resorted to the one queue and some helper functions).

If you are interested in incorporating this stuff I can include some more comments/etc, I'm just a bit busy trying to deploy this with my work atm.

Also, I'm not sure what is in the 2.0 / 2.1 release roadmap, so I guess you would have to decide. Maybe if other people also review the changes and think they are a good idea?

Cheers,
Andrew

PS: I should include a comment on the addAndDropIfNecessary() method that it is a copy of the add() method from Guava's EvictingQueue.

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -78,28 +83,60 @@ public InfluxDB setLogLevel(final LogLevel logLevel) {
...
+		return enableBatch(actions, flushInterval, 5 * flushInterval, flushIntervalTimeUnit);
+	}
+
+	@Override
+	public InfluxDB enableBatch(final int flushActions,
+			final int flushIntervalMin,
+			final int flushIntervalMax,
+			final TimeUnit flushIntervalTimeUnit) {
+		
+		enableBatch(0,
```
FlavioF: "Capacity should be > 0 or NULL", change it to null.

andrewdodd:
Despite this being very useful for me (and perhaps some other people), I think this 'auto-batching' feature is actually something that should be handled by the clients of this library.

I think the 'batch' mode should just be another write option, and we should not magically batch in the background. I think this will avoid (if not solve) a number of issues, such as:
 - what to do on failure
 - how to support async

PR #115, opened 2015-11-07
support integer type

smerschjohann:
I just saw, that the current version of influxdb-java does not support the integer type: https://influxdb.com/docs/v0.9/write_protocols/write_syntax.html

This small change will fix that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu
 MAINTAINER Stefan Majer <stefan.majer [at] gmail.com>
 
-ADD https://s3.amazonaws.com/influxdb/influxdb_0.9.1_amd64.deb /influxdb_latest_amd64.deb
+ADD https://s3.amazonaws.com/influxdb/influxdb_0.9.4.2_amd64.deb /influxdb_latest_amd64.deb
 RUN dpkg -i /influxdb_latest_amd64.deb
 
 EXPOSE 8083 8086 
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -1,5 +1,6 @@
 package org.influxdb.dto;
 
+import java.math.BigInteger;
 import java.text.NumberFormat;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


majst01:
Hi,
thanks for this PR, please update the tests to ensure all are still passing. Otherwise i will merge it.

smerschjohann:
Hi,
I've added some Unit Tests and made this feature optional.

Now you have to specify .useInteger(true) if you want to use this feature. The default behavior is to use floats as the influxdb maintainers have suggested, see: https://github.com/influxdb/influxdb/issues/3479

theoweiss:
Hi @majst01,

is there a release plan for influxdb-java? We need to have this feature for our openHAB influxdb bundle as you can see here: https://github.com/openhab/openhab/issues/2748#issuecomment-154760055 . I would prefer to use a released version of influxdb-java rather than using as self-made SNAPSHOT build.

Regards,
Theo

jazdw:
It seems like a fairly arbitrary selection of types which are stored as integer (Integer, Long and BigInteger). What about Byte or Short? I really don't think BigInteger should automatically be stored as an integer as it could hold values which wont fit in 64 bits. Require the user to convert it beforehand.

andrewdodd:
I agree with @jazdw here, I'm not convinced that there should be an 'internal default' and I agree that the user should have to convert it to the right type beforehand.

@smerschjohann, My interpretation of the default behaviour discussed at  https://github.com/influxdb/influxdb/issues/3479 is that influx will interpret numbers with a trailing 'i' as floating point numbers. I do not think that means that this client library should have a default behaviour. I can see the introduction of useInterger() is actually quite a big headache because it requires the caller to know that even if they have a fixed point number they must also call useInteger to get the right behaviour. I can see this change creating a lot of client code that looks like this:

    public List<Point> makePoints(List<Object> myValues) {
        List<Point> points = ...
        for (Object value : myValues) {
            Point point = Point.measurement("inttest,1")....
            if (value instanceof Integer || value instanceof ...) {
                point.useInteger(true);
            }
        ...
    }

Which is pretty much duplicating the checks on the inside!

smerschjohann:
@andrewdodd you don't have to know if you'll use integer types beforehand. You can simply put ".useInteger(true)" and it will use it, if your values are Integers - so no additional "instanceof" check there.

But I agree, the current selection of Integer conversions is not ideal. I will write a little bit to your other pull request.

PR #116, opened 2015-11-12
Integer type

andrewdodd:
An alternative implementation for the 'integer type' support.

As I mentioned in the comments here: https://github.com/influxdb/influxdb-java/pull/115 I think the `useInteger` behaviour is confusing (and a bit dangerous).

These changes:
 - Change the decision to be based entirely on the type of the field's `value` (i.e. via the type system only), with the upshot being: if want ints you must pass ints, if you want floats you must pass floats.

 - Change the API of the `field()` method to align exactly with the types defined for fields in: https://influxdb.com/docs/v0.9/write_protocols/line.html

 - Deprecate the type-agnostic `field(String, Object)` method, as it is the cause of this confusion. NB: I think removing it would be a good idea (NB: Most existing client code would probably not break due to type coercion).


smerschjohann:
I kind of like your solution, as you use the type system to internally decide the way it goes. But nevertheless it does not protect you at all from type mismatches. For example if you default initialize a value by zero in some routine (.field("asdf", 0) ), you will end up initializing the field with the integer type, even if you really wanted to use float types later on (programmer issue, but still a problem).

Personally, I would be fine with it, but this could brick existing code as well. I introduced the "useInteger" for exactly that reason. It will default to float-type, and only if you specify useInteger, it will append the "i" if the input value was an integer.

Maybe additionally introduce "intField()", "boolField()", "floatField", "stringField" and add Deprecated to all int/long/float field-methods?

andrewdodd:
I think I know what you are saying, but let me play it back and see if I am on the same page...

#### 1. Default initialising might cause issues
I think what you are saying here is that if you add some default value for a field, your default initialiser might have a different type to 'real' value you store against the field.

Something like:

    function Builder getDefaultBuilder() {
        Builder builder ....
        builder.field("asdf", 0);
        ....
        return builder;
    }
    ...
    Builder b = getDefaultBuilder();
    b.field("asdf", 0.0); // or b.field("asdf", false) for that matter

I really don't think this is a problem, as you should be providing the correct type for the default value anyway (i.e. you should initialise with `0` or with `0.0` appropriately). Additionally, I think it is extremely unlikely that there is a point later on in execution where you would decide, "actually I want to make anything in the fields that _could_ be an INT behave like an INT". The decision about what type you *want* is best made when you add the value.

#### 2. My change could brick existing code

I see what you mean here. In particular, you mean that it could cause code that is relying on the current implementation to coerce `int` values to floating point to "break". Updating existing deployments to include these changes would cause Influx to reject the new points, as the 'type' of the field would have changed from `float` to `int` and Influx will reject points that have a "different type than previously used".

This is a point of discussion in how much 'pain' updating client libraries should allow/enforce/avoid.

My personal view is:
 - To have this happen, you have to update the version of the `influx-java` library you are pointing too anyway; so you should probably test what happens to the rest of your code before you deploy to production;-)
 - Given that this change is about introducing `int` support, you are probably most interested in it IF you need `int`s

Either way, there is going to be pain for people who care about `int`s, either:
 - The `useInteger()` change requires:
    1. The addition of  `useInteger()` to everywhere in existing code where the field's value should be an `int`, and
    2. All future use of 'int' to include  `useInteger()`
 - This change requires:
    1. The addition of an explicit `(double)` cast to every location where the current behaviour of coercing `ints` to `floats` is still desired.

#### 3. The name of the new methods.

The new methods don't exist yet, so there is no need for adding `@deprecated` to them. However, I do not like the inclusion of the 'type' in the method name. When in doubt, I try to do what the Google Guava libraries do:
http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/base/MoreObjects.ToStringHelper.html

smerschjohann:
1) Yes I agree, you would like to enable it globally not just per point. And yes, you should initialize it by the correct value. But the Java type-safety, which a lot of people rely on, does not help here.

2) I more likely thought of the "useInteger()" as something like: "I am aware that integers could break my world", which means that the API user is aware of the pitfalls the integer types bring.
This could be dropped if we introduce the typed methods.

3) I don't get what you mean by the Guava way. The benefit of field-type-methods is that you gain the type-safety back, which we loose if we overload the methods like you did. In your guava example this is not a problem as the result name=value is always the same. This is not the case here, here we need to differentiate. 
The @deprecated-statement would just help to highlight "please use floatField() or intField()" at exactly that fields that are in doubt (everything except bool and string).

andrewdodd:
Ok, i think we are almost on the same page.

Could you provide me an example of what you mean by 1)?


Regarding 3), I'm not really sure you get the type-safety "back", as I don't think it was ever there. I can still call  `builder.doubleField(0L)` by accident. Your case is that by including the type in the method name you force the developer to think about what type to encode the data as, but they should be doing that anyway. I don't think having multiple 'typed' methods really prevents any of these errors. In fact, I think that it hurts long term maintainability (especially because these overloaded methods are on primitive types).

E.g. if I know i'm collecting a bunch of 'numeric' values, I could write a function like:

    public void buildNumericPointWithStandardTags(String fieldName, Number number) {
        ...
        builder.field(fieldName, number);
        ...
        builder.tag( with all the standard ones);
    }

But, if there are methods with the different types, I have to call the correct method on each type. Which probably results in methods like `buildIntegerPointWith...()`, `buildFloatPointWith...()` or results lots of `instanceof Float then floatField()` stuff.

In the Guava case, the example would not be the same if they had introduced `addDouble()` and `addInt()`; the API would do type coercion if it could instead of requiring the caller to organise the type they want beforehand. The good thing about overloaded methods on primitive types it they do exactly what you would expect, `int` parameters go the `int` route, `double` parameters the `double` route. You have to take care of any type casting if you want to ensure a particular method gets called.

Interestingly both of the solutions result in 'coercion' of some sort. My opinion is that if you were trying to debug a problem where an 'incorrect type' is appearing in Influx (i.e. should be an `int` but is coming through as a `float`):
  - With the overloaded methods, you know that the type system would be correctly choosing the 'int' method IF the calling type into the method was an `int`, and thus the data model must be incorrectly passing a `float` at some point.
 - With the typed methods, you have to investigate if the data model is incorrectly passing a `float` at some point (same as above) OR whether an `int` is accidentally being written via a call to `floatField()` (with the latter probably being quite hard to find).

andrewdodd:
After writing all that stuff above, I had a change of heart.

I still think the methods should be overloaded, but I have picked a different method name (`addField()`). I also included a unit test to ensure the old behaviour remains intact.

smerschjohann:
regarding 1) I meant code like this (while this code makes not much sense):

```
// get device information if online, otherwise zero
while(true) {
    if(!device_initialized) {
           point.field("temperature", 0);
    } else {
           // do real stuff here
    }

    Thread.sleep(5000);
}
```

regarding 3) You might be right, a single function is more convenient and we should just trust the programmers to do the right thing, especially if you want to do something like you explained here.

After reading your comments on this, I agree with you. If one wants to track an error down, they would simply start their debugger, which would show them the way it went - so it is okay to generally have the same function name in all cases.

With your last change of the function name, I think this solution is really the best.

macrosak:
Any plans on merging this? Would be very useful.

andrewdodd:
No idea, but I'm getting close to forking this library.

majst01:
Version  2.2 is open now and if the conclusion is this is a usefull addition im willing to merge.

majst01:
@andrewdodd can you please rebase because a merge is actually not possible.

andrewdodd:
@majst01 Yes, i'll have a look now. Any chance you got an email from my a while back?

majst01:
Hi Andrew,

Sorry for my delays but im quite bussy. I look at your email and come back
to this topic soon.

On Sun, Dec 13, 2015 at 3:28 PM, Andrew Dodd <notifications@github.com>
wrote:

> @majst01 <https://github.com/majst01> Yes, i'll have a look now. Any
> chance you got an email from my a while back?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/influxdb/influxdb-java/pull/116#issuecomment-164263188>
> .
>



-- 
Stefan Majer

andrewdodd:
No, no, it's not a problem. I was just wondering. No hurry.

andrewdodd:
@majst01 I think this PR is suitable for merging. The original author of the integer type changes agreed on Nov 12 that this change is 'best'.

PR #119, opened 2015-11-16
Batch processor keeps data points queued on failed write

PaulWeakley:
Batch processor keeps data points queued in the event that the data could not be saved to the influx database.


majst01:
Hi,

so in case of a failure of the influxdb this code will lead to a OOM on the Java Client side because the Set with unwritten points will increase unlimited.

This is not the right way to solve this problem

Issue #117, opened 2015-11-16
Ignore null values passed in PointBuilder.field methods?


scottheckel:
It seems to me that when you pass in a null value for a field, that the library should ignore this value when creating the line protocol text. The reason for this is that you could receive an error from the server "invalid number". I think this goes with [InfluxDB#2420](https://github.com/influxdb/influxdb/issues/2420).

Thoughts?

PR #129, opened 2015-12-04
Fix for issue #117

antjori:
Fixes issue #117 related to influxdb/influxdb/issues/2420.

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -286,22 +286,26 @@ public class Point {
 		numberFormat.setMinimumFractionDigits(1);
 
 		for (Entry<String, Object> field : this.fields.entrySet()) {
-			sb.append(KEY_ESCAPER.escape(field.getKey())).append("=");
 			loops++;
 			Object value = field.getValue();
-			if (value instanceof String) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/influxdb/dto/Point.java
```diff
@@ -286,22 +286,26 @@ private StringBuilder concatenateFields() {
...
-			} else if (value instanceof Number) {
-				sb.append(numberFormat.format(value));
-			} else {
-				sb.append(value);
-			}
 
-			if (loops < fieldCount) {
-				sb.append(",");
+			// Avoiding invalid null type
+			if (value != null) {
```
majst01: I think it is more readable if you continue if value == null, it also leads to less idented code. 
If you mind change it this way i will merge

antjori:
Sure, I'll do it!

src/main/java/org/influxdb/dto/Point.java
antjori: Sure, I'll do it!

antjori:
Since I didn't create a branch for this issue, I'm going to create it and do all the changes there.

Afterwards, I'll do a new pull request.

Closing this one.

Issue #130, opened 2015-12-05
It's not possible to handle numeric-named databases


slomek:
It's not possible to create (nor drop) databases which name is just a number:

    influxDB.createDatabase(numericDbName);

results in an error:

    {"error":"error parsing query: found 123, expected identifier at line 1, char 31"}

PR #131, opened 2015-12-05
Allow create/drop numeric-named databases

slomek:
This fixes https://github.com/influxdb/influxdb-java/issues/130


majst01:
If you add some tests i will merge it.

Issue #110, opened 2015-10-23
NullPointerException when providing org.influxdb.dto.Point.Builder.tag with null argument


Cyanide789:
Dear Majst01,

First of all: thanks for coding this API; it has saved me a great deal of work to integerate an ETL application with InfluxDB.

However recently, I stumbled onto a nasty NullPointerException, which you may want to address:

19:54:35.452 [task-scheduler-2] WARN  <<anonymised_package_name>>.InfluxWriter - Could not write to InfluxDB server. Underlying exception is null
java.lang.NullPointerException
        at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:210)
        at com.google.common.escape.ArrayBasedCharEscaper.escape(ArrayBasedCharEscaper.java:118)
        at org.influxdb.dto.Point.concatenatedTags(Point.java:246)
        at org.influxdb.dto.Point.lineProtocol(Point.java:236)
        at org.influxdb.dto.BatchPoints.lineProtocol(BatchPoints.java:256)
        at org.influxdb.impl.InfluxDBImpl.write(InfluxDBImpl.java:163)
        at <<anonymised_package_name>>.InfluxWriter.write(VisInfluxWriter.java:85)

I caused this (by mistake) by setting a null tag in a PointBuilder, so org.influxdb.dto.Point.Builder.tag("description", null). Now you may argue if it isn't the responsibility of the implementor to make sure this never happens, but many public APIs do null-checks and return customised error messages for it.

Just thought I'd let you know.

Kind regards,
Cyanide789

majst01:
closed by #136

PR #133, opened 2015-12-05
Fix for issue #110.

antjori:
Fix for issue #110.

Even though the map implementation used to store tags does not allow nullable keys, it is validated this possibility while concatenating tags.

If, in the future, a different map implementation is used, this issue will not reappear.

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -7,6 +7,8 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.TimeUnit;
 
+import org.influxdb.exception.NullTagException;
+
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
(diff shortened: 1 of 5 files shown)


src/main/java/org/influxdb/dto/Point.java
```diff
@@ -275,6 +278,23 @@ private StringBuilder concatenatedTags() {
...
+	 * Validates either the entry tag is null or if one of key/value is null.
+	 * 
+	 * @param tag
+	 *            the entry tag to be validated
+	 * @throws NullTagException
+	 *             if either the tag is null or if one of key/value is null
+	 */
+	private void validateTag(final Entry<String, String> tag) throws NullTagException {
+		if (tag == null) {
+			throw new NullTagException("A null tag entry was encountered! Please revew your data.");
```
macrosak: @antjori Typo: revew -> review

majst01:
This was already fixed by: #136

Issue #126, opened 2015-12-02
Allow to write "lines" directly, not just Point objects.


aadrian:
Please allow to write lines (strings) (as in the line protocol) directly, not just Point objects.

Especially when some of the data is coming from other systems already respecting the line protocol, it makes no sense to parse it, make Point objects so that the API makes them back strings and sends to the DB.

Thank you.

majst01:
Hi @aadrian 

Im happy to accept PR Requests for that.
You can start adding a new method here:

https://github.com/influxdb/influxdb-java/blob/master/src/main/java/org/influxdb/InfluxDB.java#L134

for example 
	public void write(final String lineprotocol);

and then add the implementation into:
https://github.com/influxdb/influxdb-java/blob/master/src/main/java/org/influxdb/impl/InfluxDBImpl.java#L171

Also add some Tests to verify your code works correct.

Greetings
Stefan

andrewdodd:
I question the value of this feature request.

@aadrian or @antjori or @slomek, could you supply a good example of where this functionality is really needed?

I just don't "get" it. In particular,
 1. If you already have a system that creates data in the line protocol, why are you using this library?
 2. Why not send the data directly from that system? OR
 3. If you really need this library for some reason, why not change the existing system to NOT make line protocol strings and defer that responsibility to this library?

This library provides a pretty narrow set of features. It is pretty tightly bound to the influx DB and the features it supports; and currently I think the abstractions in the library are pretty good at doing their job.

Also, by letting the `influx-java` library do the translation to the line protocol you get an additional benefit: You can pick the version of the `influx-java` library that implements the same version of the line protocol that your influxDB is using.

As an aside, this request is basically allowing a side-door, effectively an "unsafe" mode. Before it is really considered I would like some thinking in the areas of:
 - Should this function validate the provided lines?
   - Yes -> Well then you've just moved the 'need to parse' into the library. Also, do you have to supply the version of the line protocol you are using, so that the library knows how to do the parsing?
   - No -> Then the library is just a pipe.
 - Should the lines provided in this way be batched? (if the batching is enabled)
   - If not, then should they have priority over records that are being batched?
   - If yes, how should the lines be stored in the batch processor? As strings? Or parsed and stored as Point objects?

aadrian:
> I question the value of this feature request.

If I'm pumping millions of data lines very fast (in a few seconds), than the "Point" object creation overhead is just too huge.
[...]

PR #137, opened 2015-12-13
Allow to write lines directly to the database

slomek:
This fixes https://github.com/influxdb/influxdb-java/issues/126.

I've added three tests to verify that it is working correcly - any comments are welcome.

### DIFF
--- a/src/test/java/org/influxdb/InfluxDBTest.java
+++ b/src/test/java/org/influxdb/InfluxDBTest.java
@@ -1,6 +1,7 @@
 package org.influxdb;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
 import java.util.logging.Level;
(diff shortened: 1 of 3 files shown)


src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -23,177 +24,233 @@
...
  * Test the InfluxDB API.
- * 
+ *
  * @author stefan.majer [at] gmail.com
- * 
  */
 @Test
 public class InfluxDBTest {
 
-	private InfluxDB influxDB;
```
majst01: Please do not reformat code which is untouched. Can you please fix this, otherwise im fine with this PR and will merge it.

slomek: You're absolutely right, I'll fix this later today.

PR #141, opened 2015-12-29
Make BatchPoints and Point serializable, add getters to Point

RadoBuransky:


majst01:
Hi,

can you please explain what problem you try to solve ?

RadoBuransky:
Hi. I am having several problems and this is only partial solution. One problem is testing and another is serialization.

1. **Interfaces everywhere** - ideally the API to InfluxDB should contain only interfaces. That way everything can be easily mockable. Especially `BatchPoints` and `Point`. At the moment I am not able to verify call to `InfluxDB.write` because it is not possible to read values of points.

2. **Everything serializable** - I am using this library in an Apache Spark application and the way it works is that it serializes and sends data between nodes. Now it doesn't work because again `BatchPoints` and `Point` classes are not serializable.

To fulfill these two quite meaningful requirements, some breaking changes have to be introduced. Also it would be great if there are some nice convenience methods for Scala developers. I can help with this. Or even better, do it in Scala from scratch. Ha. Thanks

PR #146, opened 2016-01-16
Add support for asynchronous requests

TopShelfSolutions:


caoli5288:
May you should not bump java's version in a commit.

PR #147, opened 2016-02-17
Put should always be async

jazdw:
When using the batching functionality users do not expect the write method to be synchronous.


majst01:
@andrewdodd any objections on this PR ?

jiafu1115:
@majst01 This is one critical issue due to it prevent aysnc when size is > action . Why you don't accept it?

majst01:
@jazdw would you rebase that, i will merge it then.

PR #150, opened 2016-02-28
Fix wrong code examples

lennartkoopmann:
I *think* that there in an error in the code examples and I changed it. I am new to InfluxDB but it seems like this cannot be right.


lennartkoopmann:
Done!

Issue #149, opened 2016-02-24
Using Point.Builder is memory intensive - support plain line protocol insertion


billevans1963:
I have been using the maven central 2.1 version of influxdb-java and am finding that the Point.Builder mechanism to insert records is unnecessarily memory intensive:  Every point allocates 2 new Maps to build a single String line protocol entry.  When you are inserting hundreds of thousands of records at a time, this is not particularly efficient.

Yesterday I discovered that the latest github source contains methods to write Line Protocol directly, so I downloaded it and modified my code to write directly. It seems a much better way to go. 

Two questions:  Do you intend to develop the write(String) method further, by for example enabling  a similar kind of batching that you use for Points?  When do you plan to push the next version onto Maven Central?

andrewdodd:
Hi @billevans1963 

I guess that it a bit inefficient now that I think about it.

Could you shed any light on how you are building the line protocol strings without going via the builder?

Additionally, if you have any input on #148, especially around the implication of #126 to batching, it would be great if you could comment.

andrewdodd:
Any chance you have an example test case that demonstrates this issue? I.e. a case using Point.Builder and a case without?

billevans1963:
Hi @andrewdodd,

I replaced Point.Builder and 'enableBatching'  with the following very simple class:

`import java.util.ArrayList;
import java.util.List;

import org.influxdb.InfluxDB;
import org.influxdb.InfluxDB.ConsistencyLevel;

/**
 * InfluxBatchInserter
 * <p>NOT Thread safe.
 *
 * @author: bill.evans
 * Date: Feb 22, 2016
 */
class InfluxBatchInserter {
	//
	private final String database;
	private final String retentionPolicyName;
	private final ConsistencyLevel consistencyLevel = ConsistencyLevel.ANY;
	private final InfluxDB influxConx;
	private final int batchSize;
	//
	private List<String> linesToWrite = new ArrayList<String>();
	

	/**
	 * Constructor
	 * @param influxConx
	 * @param batchSize
	 */
	InfluxBatchInserter(InfluxDB influxConx, String database, String retentionPolicyName, int batchSize) {
		this.influxConx = influxConx;
		this.database = database;
		this.retentionPolicyName = retentionPolicyName;
		this.batchSize = batchSize;
	}
	
	/**
	 * Name: flush
	 */
	protected void flush() {
		if (linesToWrite.size() > 0) {
[...]

PR #151, opened 2016-03-01
Investigation of options for Issue #149

andrewdodd:
I've made an alternative Point class containing only the line protocol
formatted string. I've also added a crude test for determining the
difference in memory performance.

This IS NOT for inclusion, but for discussion.

See here for the object size stuff (I think): http://www.javapractices.com/topic/TopicAction.do?Id=83


majst01:
Hi @andrewdodd half year old PR, what should we do about that ?

PR #155, opened 2016-03-18
Added method to check whether batching is enabled

mariusoe:
Hi,
I've needed this method and I thought maybe it can be useful for more people :-)


majst01:
Hi,

if you add a test, i will merge it.

mariusoe:
done :)

PR #164, opened 2016-04-20
Allow non-localhost InfluxDB to be used in tests

andrewdodd:
This allows the tests to be run against an InfluxDB at any IP address. I
needed this because I am running my InfluxDB inside a Docker container
which is running inside a `docker-machine` VM.

These changes look for `INFLUXDB_IP` in the system environment for the
IP to use, falling back to localhost (127.0.0.1) if the key is not
specified.

On a Mac, from the command line do something like:

  $ export INFLUXDB_IP=www.xxx.yyy.zzz
  $ mvn test

In Eclipse, add an item to your "run configuration's" environment
variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,11 @@ Then you can build influxdb-java with all tests with:
 If you don't have Docker running locally, you can skip tests with -DskipTests flag set to true:
 
     $ mvn clean install -DskipTests=true
+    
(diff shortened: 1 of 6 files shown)


andrewdodd:
@majst01 Does this seem reasonable?

Issue #154, opened 2016-03-08
Why the library always passes timestamp to the server?


george-kozlov:
Documentation says that timestamp is optional on the server side - https://docs.influxdata.com/influxdb/v0.10/guides/writing_data/

Why is the library passing timestamp always? There is a code in the build() method of Point class that validates the incoming value and if NULL passed, it uses System.currentTimeMillis().
As the result, the library is passing timestamp to the server, not depend on user's choice.
Additional validation is in formatedTime() method of Point class.

We would like to don't pass timestamp to the server to let InfluxDB generate timestamp on its own. There are more than 100 machines on our side which will write data to the database so we would like server to guarantee uniqueness of each record.

andrewdodd:
Hi @george-kozlov 

Yes, I can see that what you are saying is true. Maybe if I get some time I can look into this.

However, for me it makes sense that the library takes the timestamps and that it does not rely on the server for the following reasons:
 - The time at which the library calls `System.currentTimeMillis()` is as close as possible (temporally) to when the event occurred (i.e. it is only a few CPU cycles between when your code created the point and when the timestamp it recorded).
 - The time is stored with the recording, so if the 'write' fails it could be retried at a later point in time but would be recorded against the correct time.
 - The effects of network latency on the recorded time would be difficult to account for and would (in my opinion) make testing/troubleshooting difficult.

On a side note, I have had success with setting up my InfluxDB server as an NTP server and having my clients sync to it (though this would obviously depend a lot on the time resolution you are after).

Additionally, I have a query about the comment 'guarantee uniqueness'... I don't think I understand this point? Are you tagging your points with something from each of the different 100 machines? (i.e. a tag like "machineId=1" / "machineId=2" or "machineMAC="aabbcc112233"?)

Thanks

george-kozlov:
Hi Andrew,

Thanks for reply.
No, we don't tag our points with any unique value, at least for this data.
Theoretically, if many machines are running, two machines can push the data with the same timestamp and similar tags, so two records will be considered as the same by the server and it won't save both.
I know that such case is the exception rather than the rule, but we already run 150 machines in the cloud, the more machine we run, the more probable the problem become.
That's why we would like to don't pass the timestamp and allow the server to generate value on its own.

In the library, API you allow the null value for the timestamp, but generate and pass timestamp in any case, no matter whether a caller is passing it or not. I believe this is wrong, and the caller should choose whether to pass the timestamp or not while working with the library.

--
Regards,
George

[...]

PR #168, opened 2016-04-28
Modified to not have time sent by default

zanella:


andrewdodd:
HI @zanella , I will have a look at this today or tomorrow. Thanks

andrewdodd:
Hi @zanella,

I'm not able to take this PR, as it changes the behaviour of the interface.

In the existing code, calling `.build()` without having set a value for the time will cause `build()` to use `.currentTimeMillis()` as it builds the `Point` object. The changes in this PR mean that instead the time will be omitted unless a call to `.time()` or `.fillTimeMillis()` is made. This will almost certainly break code that uses the current behaviour. In fact, I know that my software uses this feature (additionally, I still don't really understand the need for this change as I believe the client should be the one deciding the timestamp, but it is supported by the influx protocol so I guess it is a good inclusion).

Additionally, you haven't included any tests and would need to fix some spelling errors.

It would be great if you wanted to still add this feature, but could I request some things if you do? I'm not sure if you read the comments for #154 but I think adding a method `public Point build(boolean omitTimestamp);` would probably be the easiest way to achieve this (NB: i think the `withoutTimestamp` or `omitTimestamp` parameter name makes the most sense, also perhaps `useServerTimestamp`).

Some steps would be:
- Add a test to ensure the current behaviour is retained. I.e. something like `test_it_includes_timestamp_when_built_without_setting_time()`
- Add the test for this new feature, i.e. something like `test_it_does_not_have_a_timestamp_when_built_to_omit()`  (sorry, not a very good test name, but you can probably think of a better one)
- Retain your toString stuff (perhaps include 'time=SERVER_TO_SET' if the `time` attribute is `null`?)
- Update the readme to indicate how to use the new `build()` method

zanella:
Hey @andrewdodd ,

1) Yeah, I agree that I changed the behavior silently..., having said that, I think neither is correct, `.time()` should be deprecated so that the API change is not silent, and the decision to include a timestamp must be made by the user. About the reasoning behind the change: on my case I don't trust the hour in the client.

2) Heh, I didn't, once the code change stabilizes I'll do it. Can't see the spelling errors, can you point them out ?

3) I did, I'm **strongly** against having a special `build(boolean)`, it's just gonna make it harder for the next optional parameter, I think the best decision is not putting an optional param silently (time as it is now), but making the decision of setting it locally easy (`.fillTime<>` or some better naming scheme)

4) I'll do the tests when I get some time, but probably won't improve on the naming side, hehehe

5) Not pro or against the ` 'time=SERVER_TO_SET' `, guess I'll just do it then. Actually I think `Point` should have a `stringProtocol()` as well

- [ ] add stringProtocol()

6) Answered on 3

andrewdodd:
Hi @zanella 

Thanks for the numbers, it will make our discussion a bit easier.

1) I also agree that the user should be forced to choose the time rather than have it filled as a by-product. However, the Influxdb [docs](https://docs.influxdata.com/influxdb/v0.12/write_protocols/line/) themselves state that timestamps are optional but "_should be specified if possible_". It makes the most sense to me if the client creates the timestamp, as it is likely to get closest to the time the event occurred (i.e. no processing/network/retry latencies etc.). And in some ways the recommendation in the Influxdb docs gives me the feeling that _not including_ a timestamp should be the exceptional case that requires a specific choice by the programmer.

1) If you are going to change the behaviour of an API, I would prefer a compile time break rather than a deprecation and silent change of existing assumptions.

2) I guess it makes sense to me to make the tests before/as the code stabilises? Your function `formatedTime()` should be `formattedTime()`. Incidentally, a) wouldn't it be easier to include this functionality in the `toString()` method? b) why are you adding this to the `toString()` method at all? c) perhaps change to a verb, like `appendTimeAsNanos()`?

3) I agree that it will make the next optional parameter much harder. However, from the definitions in the Influxdb docs (as at v0.12) there are currently only 3 types of things in a point - a 'key'; the 'fields'; and the 'timestamp'. 'timestamp' is the only optional one, so I'm a bit more at ease with the `build()` option. The way this library deals with the optional 'tag' part of the key is correct. If there are more optional parts added to a point the influxdb protocol it would probably be best to restructure the `Point` class.

5) What do you mean by add `stringProtocol()`?

All great points!

zanella:
**I'll remove this line once I have addressed all the points **

1.a) I see your point..., but actually upon reading (properly) the documentation what is now bothering me is that `formattedTime()` as well as `BatchPoints.write()` and probaly others are wrong, it's always coercing the time to nanoseconds precision when it should respect the precision set by the user:

> The timestamp section is optional but should be specified if possible. The value is an integer representing nanoseconds since the epoch. If the timestamp is not provided the point will inherit the server’s local timestamp. Some write APIs allow passing a lower precision. If the API supports a lower precision, the timestamp may also be an integer epoch in microseconds, milliseconds, seconds, minutes or hours. We recommend using the least precise precision possible as this can result in significant improvements in compression.

so:

- Making it easier to do the "right thing", set the time on the client side: for this I vote on not changing the `.build()` method, people that are now doing the right thing and setting the time shouldn't have to change their code and call a `build(boolean)`

- Not create the time by default: has to be explicit w/ a runtime error like the other `Precondition.check()` calls - df09ec1

- *Respect* the precision set by the user: new MR

1.b) You're right, but I think the best we can do here is the runtime error as I said above

2.a) Working on them (tests)

2.b) Ah!, I see, it's not my func actually, I just changed the sig of the existing one, typo fixed: 860b5bb
I think I see what happened... the diff makes the `.toString()` and `lineProtocol()` seem like one, `.formattedTime()` is used by `lineProtocol()`, as I said, it's not my func, fixing the typo seems harmless since I already the sig anyway, not sure about the renaming, it really is formatting it before sending... maybe `formatTimeAsNano()` ?

andrewdodd:
First things first. I agree with all of your points that the API _should_ have been done differently originally (but hindsight is 20/20 right!).

Having read your point 1a) above I think we both agree on the following:?
 * Having the client include the timestamp is the recommended behaviour for any client library;
 * That this library does this by default when `build()` is called (which is perhaps sub-optimal); and
 * That allowing the library to _not_ specify a timestamp is probably a 'less common' usage of a library like this.

In 1a) and 1b) you mention that people currently doing the right thing should not have to change their code. I believe the 'right' thing is to include a timestamp. However, the addition of a runtime error if they don't call `fillTimeMillis()` requires them to change their code. You may as well make the change a compile time error.

1b) I don't really see the _practical_ difference between adding functions like `useServerTimestamp()` and `fillTimeMillis()`; or including `buildUseServerTimestamp()` and retaining the existing functionality for `build()`. Especially if calling one of these two functions is mandatory, and it is enforced via a runtime exception rather than a compile time error. More to this point, I don't see it as that far from `buildUseServerTimestamp()` to `build(boolean useServerTimestamp);`.

2a) Cool

2b) You are right (my mistake, I didn't read the diff correctly).

3) In 1a) above you note that the library is always coercing to nanoseconds. I agree that this is incorrect (especially given the influxdb docs suggesting using the lowest precision necessary). I think this should become another issue (however, I'm going to raise it after this issue/pr is closed).

zanella:
1a)

> I believe the 'right' thing is to include a timestamp.
Me too, that's what the MR reflects, you either set the time or call `useServerTimestamp()` explicitly

>  However, the addition of a runtime error if they don't call fillTimeMillis() requires them to change their code.
`fillTimeMillis()` is just a helper function, the user can set the time manually through `time()`. I think there should be fillTime<Precision_HERE> for all the available precisions, no need to make the user lookup how to generate the timestamp...

>  You may as well make the change a compile time error.
How? I don't think having a special 'build<BLAH>()' is a good choice.

--------------

1b) 

> I don't really see the practical difference between adding functions like useServerTimestamp() and fillTimeMillis();

Making the user decide to send the `time + precision`, or at least set the `precision`, renamed `useServerTimestamp()` to `useServerTimestampAtDefaultPrecision` (a9e3553) to make it clear.

3) Agree

> 5) What do you mean by add stringProtocol()?

The URL usable query string, `toString()' generates a debug string, and `lineProtocol()` and encoded one

andrewdodd:
HI @zanella 

Unfortunately I don't feel like you are listening to the points I've made in 1a) and 1b). I'll try summarising the issue a little:

- Including a timestamp is recommended by influxdb. Omitting a timestamp is the unusual case.
- Currently, the library always includes a timestamp. That time is 'now' by default, or some other time as provided by the user via `time()`.
- It is unfair to introduce API changes that break existing code that is already 'doing the right thing'. ('Doing the right thing' is including a timestamp, which happens to be 'now' if you don't know any better).
- Your changes require everyone to change their code, regardless of if they are doing the right thing or not:
  - If they rely on the 'now' timestamp being introduced automatically by `build()` they will have to change their code to call `fillTimeMillis()` as well. If they don't they will get a runtime exception.
  - If they are wanting to use the server's time, they have to change their code regardless of the solution. Allowing this possibility is what this issue is about. 

Considering the above points I can make the following conclusions:
 - It is unfair to introduce changes that break the existing API without good reason. For me, this set of changes doesn't present a good enough reason given their are potentially other solutions that fit the bill.
 - This requested functionality could be delivered via the `build(omitTimestamp)` mechanism or it could be delivered through something in the builder chain like `.useServerTimestamp()`. Either is fine for me, so long as it doesn't break existing code.

This has been an excellent journey for me, I think your points about adding a `build(boolean)` method have been very good (something I hadn't thought about too much). However, I think I am now even more strongly in favour of changes that do not break the API.

1b) I don't think there is any need for more helper functions. The key use cases are 'just use now' and 'use a time I provide'. I also don't think there is any need for specifying the 'millis' or the 'atDefaultPresicion' in the method names, as it over constrains them in my opinion. The methods should encapsulate the intention/request, not the details.

3) and 5) should probably be moved to their own issues.

zanella:
I see your points, but this is not a matter of misunderstanding, it's of disagreement. 

- I don't think setting default values is the right approach,

- Nor making it harder than it should be to set the time, by not having the helper functions,

- I see w/ good eyes self-explanatory method names, akin to the mantra " Explicit is better than implicit", though probably are more verbose for it.

The client lib in Go actually follows the line of thinking I'm advocating for: https://github.com/influxdata/influxdb/blob/master/client/v2/client.go#L344

3) & 5), yes, which is why I didn't put them in this MR

PR #169, opened 2016-05-03
Refactoring dependencies

FlorinAsavoaie:
The TreeMap can be used without using com.google.common.collect.Maps. There is also no need of com.google.common.collect.Ordering because the String already implements Comparable by itself and a TreeMap is already sorted.

Signed-off-by: Florin Asavoaie <florin.asavoaie@gmail.com>

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -6,12 +6,11 @@ import java.text.NumberFormat;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.TreeMap;
 import java.util.concurrent.TimeUnit;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
-import com.google.common.collect.Maps;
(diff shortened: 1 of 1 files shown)


FlorinAsavoaie:
I ended up doing these because I'm looking to implement something that could send data to telegraf using the line protocol rather than directly to InfluxDB and using the Point class from this library seamed like the proper choice. However, when compiling my code and using the maven-shade-plugin, I ended up with tons of additional code in my jar file without any reason. The change definitely does not have any negative impact, either from a functional, performance, or code readability points of view.

majst01:
But google.common.collect will be still needed after your patch, so i dont see what makes the difference.

FlorinAsavoaie:
Actually, the maven-shade-plugin will take it out during minimizeJar if it is not imported.

majst01:
And how big is the gain ?

FlorinAsavoaie:
800 KB (at least with the 2 preconditions and 1 string condition statically imported locally). It is not a big deal in size, but I like my stuff clean :).

Issue #170, opened 2016-05-05
Is accessing Point.Builder concurrently thread safe ?


Rajind:
Basically I want to do the following. I create a point builder object like this.
Point.Builder pointBuilder = Point.measurement("measurement");

Then pointBuilder.addField(), pointBuilder.addTag() will be called in multiple threads. 
I saw that in the influxdb-java source TreeMap is used which is not thread safe according to my knowledge. 

I know that TreeMap is used since the ordering is required for better performance. So can I used ConcurrentSkipListMap<>() instead ? And do we need to order fields as well ?

Rajind:
`import org.influxdb.dto.Point;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;

public class ThreadTest {
    public static Point.Builder pb;
    public static int THREADS =  1000;
    public static long averageTime = 0;

    public static void main(String args[]) throws InterruptedException {
        for (int i = 0; i < 1; i++) {
            pb = Point.measurement("m1");

            long time = System.nanoTime();
            ExecutorService service = Executors.newFixedThreadPool(THREADS);

            for (int j = 0; j < THREADS; j++) {
                final int finalJ = j;
                service.execute(new Runnable() {
                    public void run() {
                        int num = (int) Math.ceil(Math.random() * 100);
                        try {
                            Thread.sleep(THREADS - finalJ);
                        } catch (InterruptedException e) {
                            e.printStackTrace();
                        }
                        long threadId = Thread.currentThread().getId();
                        //System.out.println("Thread id: "+threadId);
                        pb.tag("tag"+threadId, "hello");
                        pb.addField("field"+threadId, threadId);
                }});
            }

            // Make sure the executor accept no new threads.
            service.shutdown();
            service.awaitTermination(Long.MAX_VALUE, TimeUnit.DAYS);
            long timeUsed = (System.nanoTime() - time) / 1000000L;
            averageTime += timeUsed;
            System.out.println("All threads are completed in "
                    + timeUsed + " ms");
        }
        System.out.println("The average time is " + averageTime / 10 + " ms");
    }
}
`

`Exception in thread "pool-1-thread-569" java.lang.NullPointerException
	at java.util.TreeMap.rotateRight(TreeMap.java:2079)
	at java.util.TreeMap.fixAfterInsertion(TreeMap.java:2111)
	at java.util.TreeMap.put(TreeMap.java:574)
[...]

PR #171, opened 2016-05-05
Using thread-safe data structures in Point class

Rajind:
Changes made to Point class. Changed tag Map to ConcurrentSkipListMap which is thread safe and also supports natural ordering. Changed tag Map to ConcurrentHashMap. (Regarding the issue #170) 
https://github.com/influxdata/influxdb-java/issues/170

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -1,19 +1,20 @@
 package org.influxdb.dto;
 
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.collect.Ordering;
+import com.google.common.escape.Escaper;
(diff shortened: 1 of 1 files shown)


majst01:
Hi,

please add a unit test to show all is still working and thread safety is given.

andrewdodd:
Hi guys,

I'm not convinced this is a necessary change. Replacing the underlying map will (may) fix the threading issues @Rajind  is facing, but everyone will pay the performance penalty. I feel like it would probably be easier for @Rajind  to put some locks in his own code.

Just my 2 cents.
A.

caoli5288:
@andrewdodd I agree:+1

majst01:
@andrewdodd god catch +1 NAK

Closing this PR

Issue #163, opened 2016-04-19
Points loss using batch processing


FlavioF:
We are using influxdb java client 2.2 (this happens in 2.1 too) and we are experiencing some data loss.

``` java

// creating the influxdb connection
InfluxDB connection = InfluxDBFactory.connect(this.influxdbHostUrl, this.influxdbUser, this.influxdbPwd) .enableBatch(5000, 30, TimeUnit.SECONDS);

//..... a lot of other stuff

// sending a lot of data 
log.info("Writing new serie {}", data);
connection.write(Constants.STATSLOG_DB_NAME,
            "default",
            Point.measurement(s.getKey())
                .time(s.getTimeInMillis(), TimeUnit.MILLISECONDS)
                .field("value", data.getMessage().getValue())
                .tag("type", data.getMessage().getType().name())
                .build());
```

In the above example, when we send like (lets say) 1k points of the same measurement, we get 1k log entries however some times we only have 999 points to that measurement in influxdb. 
There isn't any errors in influxdb log.

Can it be anything related to java client batch processing?

I am happy to help debugging the problem or event fixing it if needed.

Thank you in advance.

andrewdodd:
Yes, the current BatchProcessor implementation will silently drop batched points if they do not send successfully. Would you be willing to:
a) Test the changes from the #108 PR branch?
b) Test the master branch soon after I merge the #108 PR into the master branch?

FlavioF:
Yes, I can test it. Not problem. Tomorrow I will test the PR and let you know the result of it.

andrewdodd:
Ha! Great news. I've been hoping someone (other than me) can see if works before I commit to merging it.

FlavioF:
I run my application using your patch and there is still data loss. 
Furthermore, I added a comment to a line in InfluxDBImpl of your patch. Take a look into it.

andrewdodd:
@FlavioF thanks for the comment, you are right that the default there should be `null` instead of `0`. I will change it.

Can I enquire, are you running with a capacity limit or not on your batch processor? (I have really only used the capacity limited form, as i am uncomfortable with the unbounded capacity/capacity-based-on-JVM-memory-limits option)

FlavioF:
I run the tests without a capacity limit. Do you want me to run it with a limit?

andrewdodd:
It would be great if you could.

[...]

PR #177, opened 2016-06-18
Data Loss bugfix / Additional Features

Daniel700:
I guess that I've found the bugfix for the data loss problem:
The data loss happens if you send field values as strings which contain a backslash following with a java specific escape sequence. (See here: [Escape Sequences](http://docs.oracle.com/javase/tutorial/java/data/characters.html))
The line protocol interprets then for example `sometext\newtext` with `\n` as a new line.
That's why I have added line 357 `stringValue = stringValue.replace("\\", "/");` in Point.java as a quick fix. If you want a more sophisticated solution it's up to you. But it seems to work.

Moreover, you dont have to check the field (String) value for null in the `addField()` method (line 154 in Point.java) because you should also be able to add fields with null values. You just have to check that during the `concatenateFields()` method that you have at least one field with a value. There was also a bug in building the concatenated String if you pass null values (as a Number for example) that you had a comma at the end of the concatenated string where no comma should appear because no field has been added. 

I have also added an additional Feature to addFields on basis of a condition. Since InfluxDB is a schemaless DB you don't need always to add every field to the Point class and transmit it to the Server. Please check the tests in the `WriteTest.java` class

### DIFF
--- a/src/main/java/org/influxdb/dto/BatchPoints.java
+++ b/src/main/java/org/influxdb/dto/BatchPoints.java
@@ -1,15 +1,14 @@
 package org.influxdb.dto;
 
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import org.influxdb.InfluxDB.ConsistencyLevel;
-
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Ordering;
+import org.influxdb.InfluxDB.ConsistencyLevel;
+
+import java.util.List;
+import java.util.Map;
 
 /**
  * {Purpose of This Type}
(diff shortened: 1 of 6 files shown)


andrewdodd:
Hi @Daniel700 

Thanks for submitting a PR. 

I am not sure I fully understand what you are trying to achieve with this set of changes. I'll try to cover off the different points below:

1. The data loss that is discussed in Issue #163 is related to the batch processor not retrying buffered writes if they fail. The changes in PR #108 are an attempt to fix this issue. I don't really see how these changes fix this issue?

2. You have made changes to help with the escaping of field values. The change you made you probably make more sense if it was added to the initialisation of the `FIELD_ESCAPER` at [Point.java line 31](https://github.com/influxdata/influxdb-java/blob/master/src/main/java/org/influxdb/dto/Point.java#L31). However, I think that this is actually a concern for software that uses this library and not the library itself. For example, what if you want include a newline character into your string? With the changes in this PR it would be impossible.

3. You have made it possible to add `null` to the value in fields. I'm not really sure why anyone would want to do this?

4. You have added a number of methods to the public interface of the point builder with the form `addField(String field, something, boolean activated);`. I'm not sure I really understand these methods. I.e. why have 5 more methods just to push a boolean check inside the builder object?

As a result, I'm going to close this PR and recommend you to break the PR into smaller chunks.

Thanks,
Andrew

PR #192, opened 2016-08-25
Modify task creation method; change 'default' to 'autogen'

fandw06:
…d README


majst01:
already  fixed with #198

PR #196, opened 2016-09-08
added support for version 1.0.0 in create database query

RaphaelAudet:


pom.xml
```diff
@@ -61,6 +61,11 @@
 	</build>
 	<dependencies>
 		<dependency>
+			<groupId>org.apache.maven</groupId>
```
majst01: I would like to avoid to add additional dependencies, do you think you can create a MR without this ?

Other than that i would like to merge it.

RaphaelAudet: In order to provide the future and previous version of the influxdb (create a database with or without" IF NOT EXISTS") I did a quick version comparison using (org.apache.maven.artifact.versioning.DefaultArtifactVersion).

I could drop this package but version comparison would less elegant...
Or I can just completely drop support for previous version.

What do you prefer ?

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -202,7 +197,13 @@ public QueryResult query(final Query query, final TimeUnit timeUnit) {
 	@Override
 	public void createDatabase(final String name) {
 		Preconditions.checkArgument(!name.contains("-"), "Databasename cant contain -");
-		this.influxDBService.query(this.username, this.password, "CREATE DATABASE IF NOT EXISTS \"" + name + "\"");
+		String createDatabaseQueryString = String.format("CREATE DATABASE IF NOT EXISTS \"%s\"", name);
+		DefaultArtifactVersion versionCurrent = new DefaultArtifactVersion(this.version());
```
majst01: Wouldnt it be sufficient to check for this.version().startsWith("0.") to check for influxdb version < 1.

RaphaelAudet:
@majst01 I'm making a version without mvn. I'm also including a switch between "default" and "autogen" in the test according to the version of influxdb.

majst01:
Thanks for your contribution !

larrywalker:
How do we get a 2.3 version of influxdb-java released?  It would be great if a number of recent fixes where rolled into a new release.

PR #199, opened 2016-09-13
Added line protocol batch management into BatchProcessor class

zelig78:
I need to write in Line Protocol mode with active batch enabled
I modified the BatchProcessor class in order to manage this mode


majst01:
Hi @zelig78 

First, thanks for contributing to influxdb-java. But i have some questions and comments.

1. What is the motivation of writing a additional low level functionality ? What is missing from the current ? Performance, API usability ?
2. Describe the motivation and the problem description always in an issue first, then write a PR and with a reference
3. Try to minimize the impact on the actual codebase, do not modify any existing formattings, add javadoc for every public method and for internal classes and methods to make their purpose and contract clear
4. Add tests for all of your new code.

So in the current form this PR get closed for this reasons.

Issue #203, opened 2016-09-19
Upgrade to okhttp3/retrofit2


bentatham:
Upgrade to most recent okhttp3/retrofit2.  Luckily, they use different packages, so doesn't cause conflicts if someone is using them in their own code, but still think this should use the lastest library versions.

majst01:
Hi @bentatham 

Do you have you already a working patch ?
I wanted to do that for the next release (2.4) but hadnt enought time to do so.

Greetings

bentatham:
@majst01 Working on it now.  I'm pretty new to retrofit/okhttp, but want to learn about it...so this is a good opportunity.  A lot has changed...and not great migration docs unfortunately.

majst01:
If you have any questions, please ask i will do my best.

bentatham:
Do you really have to install and run influxdb on your local pc to run the tests?  Makes for very non-stand-alone build.

majst01:
No, depends,
look at compile-and-test.sh , as long as you have docker installed on your machine no additional installations are required to run all tests.

bentatham:
sorry about all the squashed commits -- didn't realize they would end up in the list above.  I'm relatively new to using github like this.

PR #204, opened 2016-09-20
Issue #203: use okhttp3/retrofit2/moshi

bentatham:
Here is an update from old okhttp/retrofit to okhttp3/retrofit2/moshi.  I chose moshi (over, say, gson) since it is my understanding that since is also uses okio, it is the most memory efficient of the converters.

Logging moved from retrofit to okhttp, using an interceptor.

Error handling has changed too, and is now directly handled by exceptions instead.

I can't run docker on my pc right now due to the need for virtualbox for some other dev I'm doing, but installed influxed locally and it all passes.  The one change might be the user/pass combo - not sure if the docker container has different default passwords for some reason.


majst01:
Excellent !
I will take some time to review, but so far looks good.

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -175,59 +198,73 @@ public void write(final String database, final String retentionPolicy, final Con
 	 * {@inheritDoc}
 	 */
 	@Override
-	public QueryResult query(final Query query) {
-		QueryResult response = this.influxDBService
-				.query(this.username, this.password, query.getDatabase(), query.getCommand());
-		return response;
+	public QueryResult query(final Query query)	{
+		return execute(query.requiresPost()
+				? this.influxDBService.postQuery(this.username, this.password, query.getDatabase(), query.getCommand())
```
majst01: Please do not use ternary form, convert to more readable if then else form.

majst01:
We should also add a notice in the Readme that beginning with 2.4 retrofit2/okhttp3 is used and the api usage differs

bentatham:
Given the API diff, should it be 3.0?  I defer to you on that of course.

majst01:
Hi @bentatham Thanks for your work, i think this is ready to be merged now but i would get feedback from my co-maintainer as well.

@andrewdodd  Can you please have a look at the MR and comment.

majst01:
As Andrew did not respond i am merging this now.

PR #205, opened 2016-09-21
Fixed typo in error message

kay:
This is a trivial fix to improve the e rror message


src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -196,7 +196,7 @@ public QueryResult query(final Query query, final TimeUnit timeUnit) {
 	 */
 	@Override
 	public void createDatabase(final String name) {
-		Preconditions.checkArgument(!name.contains("-"), "Databasename cant contain -");
```
bentatham: wished we fixed the ``cant`` too...``can't`` or ``cannot``.

Issue #217, opened 2016-10-31
add conversion from unix-time to influxdb timestamp string


noamBarkai:
we use the following code in our queries of influxdb (code below is in Scala, and uses JodaTime, can easily convert to Java if needed). Thought of adding it as some helper function to the repo. 
Can also do the reverse formatting if needed (might be relevant to #190?).

```
private def influxDbTimeFormat(time: Long) = {
    // influx db timestamps are like this: 2016-09-04T17:51:26.883Z  (UTC)
    val date = new DateTime(time, DateTimeZone.UTC)
    DateTimeFormat.forPattern("yyyy-MM-dd").print(date) + "T" + DateTimeFormat.forPattern("HH:mm:ss.SSS").print(date) + "Z"
  }
```
we use it like so:
```
influxDB.query(new Query(s"SELECT * FROM some_measurement WHERE time >= '${influxDbTimeFormat(timeStart)}' AND time <= '${influxDbTimeFormat(timeEnd)}'", some_database))
```

majst01:
Like the idea, if you would like to write a MR i would accept it.

PR #218, opened 2016-10-31
datetime util functions

noamBarkai:
created util functions following issue #217 
implementation only uses Java 7 internal date-time functions. it could have been shorter if JodaTime or Java 8 were used, but I didn't want to change any dependencies / project settings.


majst01:
This was fast, thanks for contributing.

fransflippo:
Just to double-check: there's no way to use this function directly (inline) in a query? Like `SELECT toInfluxDBTimeFormat(my_epoch_field) FROM ...` ?

PR #221, opened 2016-11-02
support creating instance without username and password

jiafu1115:
refer to influxdb offical doc: https://docs.influxdata.com/influxdb/v0.9/administration/authentication_and_authorization/#set-up-authentication
`Authorization is only enforced once you’ve enabled authentication. By default, authentication is disabled, all credentials are silently ignored, and all users have all privileges.`
 So there is no need to set username and password for influxdb which doesn't enable authorization.

### DIFF
--- a/src/main/java/org/influxdb/InfluxDBFactory.java
+++ b/src/main/java/org/influxdb/InfluxDBFactory.java
@@ -15,7 +15,21 @@ import okhttp3.OkHttpClient;
  * 
  */
 public enum InfluxDBFactory {
+	
 	;
+	
+	/**
+	 * Create a connection to a InfluxDB.
+	 * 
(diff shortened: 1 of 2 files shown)


jiafu1115:
@majst01 Can you take a look at this PR too? Thanks.

majst01:
Hi,

looks good, tests are missing. Please add unit-test and i will merge.

jiafu1115:
@majst01 Ok, NP, 3ks

jiafu1115:
@majst01 done with unit-tests added. Thanks for your review again.

jiafu1115:
@majst01  Thanks.BTW. Can I create one PR for the UDP support of java client?

majst01:
yes

Issue #157, opened 2016-03-26
support UDP Protocol


gunlee01:
I hope it to support line protocol on UDP for some specific light-weight processing.

bentatham:
@majst01 - would there be an appetite to add this, if I provide a pull request?  

I was thinking of a new subclass of ``InfluxDB`` that throws ``UnsupportedOperationException`` for those methods that require a response.  Alternative approach: extract the ``write`` methods to a super-interface called ``PointsInfluxDB`` or ``WriteOnlyInfluxDB`` or something like that.    Thoughts?

mhiller:
+1 to this. . I would rather not role my own if i dont have to.

jiafu1115:
refer to #226  I am focus on it. Welcome any advices.

PR #226, opened 2016-11-04
support udp protocol

jiafu1115:
This PR try to support udp protocol so that it can fix the issue:  #157 

I just complete the draft without unit-test and without improve the udp socket send logical.
So can you help to take a fast look at to check if overall design is OK without consider UDP socket send part which I will improve it this evening.

### DIFF
--- a/src/main/java/org/influxdb/dto/AbstractBatchPoints.java
+++ b/src/main/java/org/influxdb/dto/AbstractBatchPoints.java
@@ -1,0 +1,141 @@
+package org.influxdb.dto;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Ordering;
+
+/**
+ * {Purpose of This Type}
+ *
+ * {Other Notes Relating to This Type (Optional)}
+ *
+ * @author stefan
+ *
+ */
+public abstract class AbstractBatchPoints<T extends AbstractBatchPoints<T>> {
+	protected Map<String, String> tags;
+	protected List<Point> points;
+
+	AbstractBatchPoints() {
+		// Only visible in the Builder
+	}
+ 
+	/**
+	 * The Builder to create a new BatchPoints instance.
+	 */
+	public abstract static class Builder<T extends Builder<T>>{
+		protected final Map<String, String> tags = Maps.newTreeMap(Ordering.natural());
+		protected final List<Point> points = Lists.newArrayList();
+
+
+		/**
+		 * Add a tag to this set of points.
+		 *
+		 * @param tagName
+		 *            the tag name
+		 * @param value
+		 *            the tag value
+		 * @return the Builder instance.
+		 */
+		public T tag(final String tagName, final String value) {
+			this.tags.put(tagName, value);
+			return (T) this;
+		}
+
+		/**
+		 * Add a Point to this set of points.
+		 *
+		 * @param pointToAdd
+		 * @return the Builder instance
+		 */
+		public T point(final Point pointToAdd) {
+			this.points.add(pointToAdd);
+			return (T) this;
+		}
+
+		/**
+		 * Add a set of Points to this set of points.
+		 *
+		 * @param pointsToAdd
+		 * @return the Builder instance
+		 */
+		public T points(final Point... pointsToAdd) {
+			this.points.addAll(Arrays.asList(pointsToAdd));
+			return (T) this;
+		}
+
+	}
+
+	/**
+	 * @return the points
+	 */
+	public List<Point> getPoints() {
+		return this.points;
+	}
+
+	/**
+	 * @param points
+	 *            the points to set
+	 */
+	void setPoints(final List<Point> points) {
+		this.points = points;
+	}
+
+	/**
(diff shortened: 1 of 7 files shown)


jiafu1115:
Can you help to take a look at overral design to check if it is ok so that I can go ahead with this design.Thanks

majst01:
After a first look this seems to be to many changes just to add udp support (~600 lines changed)
I see your point with the abstraction introduced but i would feel much more comfortable if the first shot of udp support is a smaller change.

jiafu1115:
Thanks for your advice. I will figure out if we can use another style or methods for it so that the changes is smaller. Thus, if we want to support batch send with udp we can't reduce too much code changes.
if only support sync send with udp. it may not involved too many changes.
Whatever, I should complete this PR if the soluction is ok except code change is big.

src/main/java/org/influxdb/dto/AbstractBatchPoints.java
```diff
@@ -0,0 +1,141 @@
...
+		 *
+		 * @param tagName
+		 *            the tag name
+		 * @param value
+		 *            the tag value
+		 * @return the Builder instance.
+		 */
+		public T tag(final String tagName, final String value) {
+			this.tags.put(tagName, value);
+			return (T) this;
```
jiafu1115: if any other method to avoid cast?

jiafu1115:
@majst01 I had completed all code to support UDP protocal. To support batch send with UDP in asyn mode and can't influence old logical of http. I failed to reduce code changes. So can you review the code if you can ignore the code changes number? Thanks.

I notice influx-python and influx-ruby, etc also support UDP, so it is important feature we should support to improve the performance.
To avoid conflict with newest master again and again. Can we take look at this PR ASAP?

Thanks.

majst01:
Still 600 lines changed ?

jiafu1115:
Yes. I compare the implements with python and ruby. The current design is more better than them due to we support async udp and support multi udp port per database. What's more, when enable udp we also can use old api for http and no need to pass extra parameters such as database name which isn't needed by udp.  So it is better design than python though more code changes. Can we review it no matter with code changes number? Thanks

jiafu1115:
@majst01  WDYT? Thanks

jiafu1115:
@majst01  Can you help to enable UDP in influxdb so that I can add tests for this PR.

Change enabled to true
Set udp support database
set bind-address.
etc.
[[udp]]
  enabled = false  
  # bind-address = ""
  # database = "udp"
  # retention-policy = ""

  # These next lines control how batching works. You should have this enabled
  # otherwise you could get dropped metrics or poor performance. Batching
  # will buffer points in memory if you have many coming in.

  # batch-size = 1000 # will flush if this many points get buffered
  # batch-pending = 5 # number of batches that may be pending in memory
  # batch-timeout = "1s" # will flush at least this often even if we haven't hit buffer limit
  # read-buffer = 0 # UDP Read buffer size, 0 means OS default. UDP listener will fail if set above OS max.

  # set the expected UDP payload size; lower values tend to yield better performance, default is max UDP size 65536
  # udp-payload-size = 65536


Many thanks!

jiafu1115:
@majst01 Can you help to enable UDP so that I can test UDP support, Thanks

enable configure can followed this configure:

[[udp]]
  enabled = true
  bind-address = ":8089"
  database = "udp"
  retention-policy = ""

  # These next lines control how batching works. You should have this enabled
  # otherwise you could get dropped metrics or poor performance. Batching
  # will buffer points in memory if you have many coming in.

  batch-size = 1000 # will flush if this many points get buffered
  batch-pending = 5 # number of batches that may be pending in memory
  batch-timeout = "1s" # will flush at least this often even if we haven't hit buffer limit
  read-buffer = 0 # UDP Read buffer size, 0 means OS default. UDP listener will fail if set above OS max.

then make you use the changed configure to run such as :
nohup ./influxd  -config youinfluxdbpath/etc/influxdb/influxdb.conf run  &

majst01:
You have to create a influxdb.conf with this settings enabled and pass it as effective configuration to the influxdb docker container and make it open 8089/udp like so:

*from compile-and-test.sh*
```
docker run -d --name influxdb -p 8086:8086 -p 8089:8089/udp -v ${PWD}/influxdb.conf:/etc/influxdb/influxdb.conf influxdb:${version}
```

If your unit-test now point to influxdb:8089/udp they should work

jiafu1115:
so do you mean that it need me to add the configure by myself? Do I have privilege？

majst01:
@jiafu1115 I can help you, give me some time to enhance *compile-and-test.sh* to make influxdb listen on udp as well.

jiafu1115:
@majst01 OK, Thanks. I  try to create another PR for try, you can help to edit the PR. Thanks

jiafu1115:
@majst01 **OK, When you finish docker file support UDP , notify me.** Then I can add test it.
BTW: I may only support UDP step by step to avoid too much code changes. For example, we can support sync send with UDP as first. This will make us more comfortable.

majst01:
Now you can start adding upd related code which can be integration tested against the influxdb container.

jiafu1115:
@majst01  Thanks. Let's support sync mode writing by UDP as first!

majst01:
I' closing this no

PR #229, opened 2016-11-05
fix warning: no need to use Throwables.propagate when we had known it is IOException

jiafu1115:


majst01:
This will be an API Change, any reason why do so ?

jiafu1115:
Yes. It is warning for deprecated and I look into the guava implement. When catching io exception. It will throw runtime exception as I changed.So we doesn't need to call throws class when we had know it is io exception. So the change will remove the warning and reduce the process in guava: check if it is unchecked exception

jiafu1115:
@majst01  you can take a look at this PR now when you are free.

PR #230, opened 2016-11-05
correct preconditions checks for primary type

jiafu1115:


majst01:
This is also a API Change, unit-test is missing

jiafu1115:
Thanks.  I will add the unit tests

PR #231, opened 2016-11-06
support setting thread factory for batch processor

jiafu1115:
[Motivation]

batch processor's thread can't be set with name or isDamen and so on. It is hard to figure out which thread is for influxdb and take some customized actions.

[Modifications]
add thread factory into processor's builder class and all another method for enable batch.

[Result]
the batch processor's thread can be customized.

### DIFF
--- a/src/main/java/org/influxdb/InfluxDB.java
+++ b/src/main/java/org/influxdb/InfluxDB.java
@@ -1,6 +1,7 @@
 package org.influxdb;
 
 import java.util.List;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
 import org.influxdb.dto.BatchPoints;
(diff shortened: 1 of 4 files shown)


majst01:
Hi @jiafu1115 

This pull request does:
- not have any description what is the purpose or a description what problem it solves,
- not pass unit tests

so i cant merge it

jiafu1115:
@majst01  I am fixing the unit tests and the PR is similar with #79. Thus, #79 is out of date (one year ago) and conflict now.

jiafu1115:
the purpose or a description what problem it solves,
we need know the thread name for trouble shooting so we need we can set it by thread factory
we need set the thread to damen to auto close the batch processor
we need other function which can involved by thread factory.

jiafu1115:
@majst01 done with fix build error and missed description of this PR! 
Thanks!

jiafu1115:
@majst01 done with fix build error and missed description of this PR! 
Thanks!

majst01:
looks good now.

PR #233, opened 2016-11-08
remove duplicated check for time is null

jiafu1115:
in builder's build, it had checked the time is null so that the time for point never be null. So it can be removed to avoid duplicated check and improve the code coverage.


majst01:
good catch

PR #234, opened 2016-11-08
Create influxdb.conf with UDP enable

jiafu1115:


majst01:
@jiafu1115 almost :-) But to much config and the given Dockerfile is not used, i close this PR and will to the needed. 

Thanks

PR #236, opened 2016-11-08
fix sending mutiple queries with single query bug

shanexu:
should use ```@Filed``` and ```@FormUrlEncoded```, or the arguments will encoded as query string append to request URL.

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -49,6 +49,8 @@ import java.util.concurrent.atomic.AtomicLong;
 public class InfluxDBImpl implements InfluxDB {
   static final okhttp3.MediaType MEDIA_TYPE_STRING = MediaType.parse("text/plain");
 
+  private static final String SHOW_DATABASE_COMMAND_ENCODED = Query.encode("SHOW DATABASES");
+
   private final InetAddress hostAddress;
   private final String username;
   private final String password;
@@ -316,10 +318,10 @@ public class InfluxDBImpl implements InfluxDB {
     Call<QueryResult> call;
     if (query.requiresPost()) {
       call = this.influxDBService.postQuery(this.username,
-                                            this.password, query.getDatabase(), query.getCommand());
+                                            this.password, query.getDatabase(), query.getCommandWithUrlEncoded());
     } else {
       call = this.influxDBService.query(this.username,
-                                        this.password, query.getDatabase(), query.getCommand());
+                                        this.password, query.getDatabase(), query.getCommandWithUrlEncoded());
     }
     return execute(call);
   }
@@ -330,7 +332,7 @@ public class InfluxDBImpl implements InfluxDB {
   @Override
   public QueryResult query(final Query query, final TimeUnit timeUnit) {
     return execute(this.influxDBService.query(this.username, this.password, query.getDatabase(),
-        TimeUtil.toTimePrecision(timeUnit), query.getCommand()));
+        TimeUtil.toTimePrecision(timeUnit), query.getCommandWithUrlEncoded()));
   }
 
   /**
@@ -343,7 +345,7 @@ public class InfluxDBImpl implements InfluxDB {
     if (this.version().startsWith("0.")) {
       createDatabaseQueryString = String.format("CREATE DATABASE IF NOT EXISTS \"%s\"", name);
     }
-    execute(this.influxDBService.postQuery(this.username, this.password, createDatabaseQueryString));
+    execute(this.influxDBService.postQuery(this.username, this.password, Query.encode(createDatabaseQueryString)));
   }
 
   /**
@@ -351,7 +353,8 @@ public class InfluxDBImpl implements InfluxDB {
    */
   @Override
   public void deleteDatabase(final String name) {
-    execute(this.influxDBService.postQuery(this.username, this.password, "DROP DATABASE \"" + name + "\""));
+    execute(this.influxDBService.postQuery(this.username, this.password,
+                                           Query.encode("DROP DATABASE \"" + name + "\"")));
   }
 
   /**
@@ -360,7 +363,7 @@ public class InfluxDBImpl implements InfluxDB {
   @Override
   public List<String> describeDatabases() {
     QueryResult result = execute(this.influxDBService.query(this.username,
-                                                            this.password, "SHOW DATABASES"));
+                                                            this.password, SHOW_DATABASE_COMMAND_ENCODED));
     // {"results":[{"series":[{"name":"databases","columns":["name"],"values":[["mydb"]]}]}]}
     // Series [name=databases, columns=[name], values=[[mydb], [unittest_1433605300968]]]
     List<List<Object>> databaseNames = result.getResults().get(0).getSeries().get(0).getValues();
--- a/src/main/java/org/influxdb/dto/Query.java
+++ b/src/main/java/org/influxdb/dto/Query.java
@@ -1,5 +1,10 @@
 package org.influxdb.dto;
 
+import com.google.common.base.Charsets;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+
 /**
  * Represents a Query against Influxdb.
  *
@@ -37,6 +42,13 @@ public class Query {
    */
   public String getCommand() {
     return this.command;
+  }
+
+  /**
+   * @return url encoded command
+   */
+  public String getCommandWithUrlEncoded() {
+    return encode(this.command);
   }
 
   /**
@@ -80,5 +92,19 @@ public class Query {
     } else if (!database.equals(other.database))
       return false;
     return true;
+  }
+
+  /**
+   * Encode a command into {@code x-www-form-urlencoded} format.
+   * @param command
+   *            the command to be encoded.
+   * @return a encoded command.
+   */
+  public static String encode(String command) {
+    try {
+      return URLEncoder.encode(command, Charsets.UTF_8.toString());
+    } catch (UnsupportedEncodingException e) {
+      throw new RuntimeException(e);
+    }
   }
 }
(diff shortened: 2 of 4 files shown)


majst01:
I already had the same idea, thanks, would you please show the behavior in a unit-test then i will merge.

shanexu:
```java
    @Test
	public void testPostQuery() {
		this.influxDB.query(new Query("CREATE DATABASE mydb", "mydb"));
		QueryResult result = this.influxDB.query(new Query("SELECT * FROM not_exists;SELECT * FROM not_exists", "mydb", true));
		Assert.assertEquals(2, result.getResults().size());
	}
```

this will fail if use ```@Query```

majst01:
Yes of course, but include some tests, one for failing and one for working which shows how your code modifications make it work. A note on the README.md regarding multiple Queries in one request would also be helpful.

shanexu:
``` java
    @Test
	public void testPostQuery() {
		OkHttpClient.Builder builder = new OkHttpClient.Builder();
		builder.readTimeout(1, TimeUnit.MINUTES);
		builder.connectTimeout(1, TimeUnit.MINUTES);
		builder.writeTimeout(1, TimeUnit.MINUTES);
		this.influxDB = InfluxDBFactory.connect("http://" + TestUtils.getInfluxIP() + ":" + TestUtils.getInfluxPORT(true), "admin", "admin", builder);
		String dbName = "unittest_" + System.currentTimeMillis();
		this.influxDB.createDatabase(dbName);
		String query = "SELECT * FROM not_exists";
		StringBuilder sb = new StringBuilder();
		for (int i = 0; i < 1 << 18; i++) {
			sb.append(query).append(";");
		}
		sb.deleteCharAt(sb.length() - 1);
		this.influxDB.query(new Query(sb.toString(), dbName, true));
		this.influxDB.deleteDatabase(dbName);
	}
```
this test case should throw exception if use ```@Query```, http header is limited to a fixed size. When using ```@Query```, the arguments will append to request url, and create a http request with large header.
When using `@Field`, the arugments are encoded into form and append to http request body.

majst01:
@shanexu can you please rebase and resolve the merge conflict ?
After that i would like to merge it

majst01:
@shanexu You should rebase, or clone again. The way you did it now does not show what you changed.

Please try again

shanexu:
Now, using ```@Field``` will cause 
```java
    @Test
    public void testWriteEnableGzip() {
```
this unit test case fail.
I need to read influxdb's source code to verify the problem.

shanexu:
@majst01 is it OK now?

jiafu1115:
@shanexu you can squash your commit into one commit to make PR's commit clean:
now your PR had 6 commits.Just one suggestion you can ignore it. Thanks.

majst01:
@jiafu1115 what do you think, i am fine with this PR

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -99,6 +99,24 @@ public void testQuery() {
...
+		this.influxDB.deleteDatabase(dbName);
+	}
+
+	@Test
+	public void testPostQuery() {
+		String dbName = "unittest_" + System.currentTimeMillis();
+		this.influxDB.createDatabase(dbName);
+		QueryResult result = this.influxDB.query(new Query("SELECT * FROM not_exists;SELECT * FROM not_exists", dbName, false));
+		Assert.assertEquals(2, result.getResults().size());
+		this.influxDB.deleteDatabase(dbName);
```
jiafu1115: try...finally to make use tmp database to be removed

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -99,6 +99,24 @@ public void testQuery() {
...
+	public void testMultipleQuery() {
+		String dbName = "unittest_" + System.currentTimeMillis();
+		this.influxDB.createDatabase(dbName);
+		QueryResult result = this.influxDB.query(new Query("SELECT * FROM not_exists;SELECT * FROM not_exists", dbName));
+		Assert.assertEquals(2, result.getResults().size());
+		this.influxDB.deleteDatabase(dbName);
+	}
+
+	@Test
+	public void testPostQuery() {
```
jiafu1115: add java doc or change test's method's name to show what's your test focus?

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -99,6 +99,24 @@ public void testQuery() {
 		this.influxDB.query(new Query("DROP DATABASE mydb2", "mydb"));
 	}
 
+	@Test
+	public void testMultipleQuery() {
+		String dbName = "unittest_" + System.currentTimeMillis();
+		this.influxDB.createDatabase(dbName);
+		QueryResult result = this.influxDB.query(new Query("SELECT * FROM not_exists;SELECT * FROM not_exists", dbName));
+		Assert.assertEquals(2, result.getResults().size());
+		this.influxDB.deleteDatabase(dbName);
```
jiafu1115: should using try...finally to delete tmp db

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -100,6 +100,37 @@ public void testQuery() {
...
+	}
+
+	/**
+	 * Test multiple queries with an HTTP POST request return results having the same size of queries.
+	 */
+	@Test
+	public void testPostQuery() {
+		String dbName = "unittest_" + System.currentTimeMillis();
+		try {
+			this.influxDB.createDatabase(dbName);
```
shanexu: I don't think it's necessary. Droping a non-existent database will not cause any problem.

jiafu1115: it is one of best practices instead of if it is necessary for this case. So it is up to you. just one advice for best practice.

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -318,8 +319,9 @@ public QueryResult query(final Query query) {
       call = this.influxDBService.postQuery(this.username,
                                             this.password, query.getDatabase(), query.getCommand());
     } else {
+      String encodedCommand = UrlEscapers.urlFormParameterEscaper().escape(query.getCommand());
```
jiafu1115: why need this (I notice the encoded=true in all query methods)? if need it, why loss it in write method on 333 line?

shanexu: okhttp-3.4.2 use different policies on encoding query parameter and form field.

For query parameter
[HttpUrl.java#L1120](https://github.com/square/okhttp/blob/parent-3.4.2/okhttp/src/main/java/okhttp3/HttpUrl.java#L1120)
```java
    /** Encodes the query parameter using UTF-8 and adds it to this URL's query string. */
    public Builder addQueryParameter(String name, String value) {
      if (name == null) throw new NullPointerException("name == null");
      if (encodedQueryNamesAndValues == null) encodedQueryNamesAndValues = new ArrayList<>();
      encodedQueryNamesAndValues.add(
          canonicalize(name, QUERY_COMPONENT_ENCODE_SET, false, false, true, true));
      encodedQueryNamesAndValues.add(value != null
          ? canonicalize(value, QUERY_COMPONENT_ENCODE_SET, false, false, true, true)
          : null);
      return this;
    }
```
it does not encode some characters, for example ```;```, but influxdb need ```;``` encode into ```%3B```.

For form field
[FormBody.java#L110](https://github.com/square/okhttp/blob/parent-3.4.2/okhttp/src/main/java/okhttp3/FormBody.java#L110)
```java
    public Builder add(String name, String value) {
      names.add(HttpUrl.canonicalize(name, FORM_ENCODE_SET, false, false, true, true));
      values.add(HttpUrl.canonicalize(value, FORM_ENCODE_SET, false, false, true, true));
      return this;
    }
```
it does encode ```;``` and some other characters.

jiafu1115: So why loss encode in line 333?  It  is  query too

shanexu: my fault, I see, i missed this one. i will fix this

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -362,7 +362,7 @@ public void deleteDatabase(final String name) {
   @Override
   public List<String> describeDatabases() {
     QueryResult result = execute(this.influxDBService.query(this.username,
-                                                            this.password, "SHOW DATABASES"));
+                                                            this.password, "SHOW+DATABASES"));
```
jiafu1115: SHOW+DATABASES? looks strange.....

shanexu: a space can be encoded into ```+``` or ```%2B```

jiafu1115:
@shanexu  I have one idea that if we can move url encode into query.getCommand()? Then we may limit the code change lines and not change every where. For known command, we can does't encode it such as "create database" WDYT?

jiafu1115:
@shanexu do you have qq? we can talk offline.

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

jiafu1115:
LGTM now. you missed the test cases so that coverage decreased

majst01:
I`m going to merge this now, thanks for all of you for digging into that issue.

PR #237, opened 2016-11-10
correct exception type when enable batch again

jiafu1115:
IllegalArgumentException is not right for this case.


jiafu1115:
squash error so I close it

PR #239, opened 2016-11-10
prepare code changes to support UDP async

jiafu1115:
This is another PR to prepare supporting UDP async. 
Create this PR's goal to make sure we can step by step with limit code changes for clear review.

abstract batch entry to support both http batch entry and udp batch entry.

This will avoid use the same batch entry due to http batch entry doesn't  need udp port and udp patch doesn't need dbname or rp.

I doesn't add any unit tests(so decrease coverage by 0.42%) due to just refactor and no use the **UdpBatchEntry** just now. it is just one common java bean without any logical.
I will add tests together with the next PR when support UDP patch officially.

### DIFF
--- a/src/main/java/org/influxdb/impl/BatchProcessor.java
+++ b/src/main/java/org/influxdb/impl/BatchProcessor.java
@@ -29,7 +29,7 @@ import com.google.common.collect.Maps;
 public class BatchProcessor {
 
   private static final Logger LOG = Logger.getLogger(BatchProcessor.class.getName());
-  protected final BlockingQueue<BatchEntry> queue = new LinkedBlockingQueue<>();
+  protected final BlockingQueue<AbstractBatchEntry> queue = new LinkedBlockingQueue<>();
   private final ScheduledExecutorService scheduler;
   final InfluxDBImpl influxDB;
   final int actions;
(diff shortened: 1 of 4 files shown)


majst01:
Build fails !

jiafu1115:
@majst01 build fail is not caused by checkstyle or test but code covearge decrease coverage by 0.42%. So you can ignore it. Due to " just refactor and no use the UdpBatchEntry just now. it is just one common java bean without any logic.
I will add tests together with the next PR when support UDP patch officially."

majst01:
No one unit-test is  failing:
see https://travis-ci.org/influxdata/influxdb-java/builds/174660265#L4749

jiafu1115:
@majst01  build fail is caused by sometime UDP's write executed on server is after http's query. so I may sleep 2s to avoid the fail.

jiafu1115:
@majst01  It is ok now. You can go ahead with your review.  If the similar build error happen again, I may change the design for the test cases to avoid sleeping 2s but just try times with interval.

majst01:
I'm still not sure why we need batchprocessing for udp because udp is fire and forget anyway.

Can you please provide some metrics which show the gains of using batchprocessing for udp.
This results should also be added to some documentation.

jiafu1115:
@majst01 
batch processing's goal is to avoid pass same tags into every points. There is no relationships for performance. It will make api user more easier to use it:

such as : if we not support batch, we may write code as followed:

```
point one = point.tag("sametagkey1","sametagkey1").tag("sametagkey2","sametagkey2").field("1","2").build
point two = point.tag("sametagkey1","sametagkey1").tag("sametagkey2","sametagkey2").field("3","4").build

influxdb.write(udpport, Arrays.asList(one, two))
```

the if we support batch:

```
UdpPatch udpPatch =UdpPatch.tag("sametagkey1","sametagkey1").tag("sametagkey2","sametagkey2").build;
point one = point.field("1","2").build
point two = point.field("3","4").build

influxdb.write(udpport, udpPatch )
```

there is another reason we should support it:  http support it. we should keep udp support too.

What's more, this PR both prepare batch and aysnc support. not just limit to batch. may be I need to change title.  Thanks

WDYT?

jiafu1115:
@majst01  Can you continue this PR just now so that I can complete support UDP async one/two days. Then we will complete all big feature such as UDP and Gzip. Then we can take a rest.

Issue #180, opened 2016-06-21
Compress large batch points write requests using gzip


cyrilmhansen:
I am currently prototyping the migration from a RRD4J System to InfluxDB. I have archive of data in CSV format, and I am writing the import tool to InfluxDB using influxdb-java.

I used wireshark to look at my network traffic, and it is quite large. I checked with curl and I  see that InfluxDB 0.13 support gzip compression of POST request body. 

As the server supports it, would it be possible to provide an option to compress large write requests with influxdb-java ?

If found this message which may be technically related : https://github.com/square/okhttp/wiki/Interceptors#rewriting-requests

Regards,

Cyril H

madhukard:
+1

viragkothari:
+1

har-ki:
+1

majst01:
Hi,

once #203 is merged this would be an option. If someone wants to show how that would look like, i happily accept pull requests.

majst01:
anyone with a working PR ?

jiafu1115:
@majst01 I can resolve this issue after I complete UDP's PRs.

jiafu1115:
@cyrilmhansen @madhukard @viragkothari @har-ki  @majst01  support it on #241, let's waiting for @majst01 's review and feedback.

PR #241, opened 2016-11-11
support gzip for http request to improve performance

jiafu1115:
as titled:

as I tested:
we can compare the content length be reduced from 2121 to 480. **it is big improvement for write's performance** though it may take more cpu time:

final_gzip_compare

and the content be compressed to gzip:
final_gzip_1

The original content is as followed:
```
                    "cpu,atag=test1 idle=100,usertime=10,system=1",
                    "cpu,atag=test2 idle=200,usertime=20,system=2",
                    "cpu,atag=test3 idle=300,usertime=30,system=3"
...............................more lines------------------------
```

@majst01 Thanks for your review!

### DIFF
--- a/src/main/java/org/influxdb/impl/GzipRequestInterceptor.java
+++ b/src/main/java/org/influxdb/impl/GzipRequestInterceptor.java
@@ -1,0 +1,76 @@
+package org.influxdb.impl;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import okhttp3.Interceptor;
+import okhttp3.MediaType;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import okio.BufferedSink;
+import okio.GzipSink;
+import okio.Okio;
+
+/**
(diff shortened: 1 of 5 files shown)


src/main/java/org/influxdb/impl/GzipRequestInterceptor.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public void disable() {
+        enabled.set(false);
+    }
+
+    @Override
+    public Response intercept(Interceptor.Chain chain) throws IOException {
+        if (!enabled.get())
+            return chain.proceed(chain.request());
```
majst01: Missing brackets

jiafu1115: @majst01  done and I squash commit to one to make commit clean. You can merge it now the close the issue #180 . Thanks again.

jiafu1115:
@majst01 done and I squash commit to one to make commit clean. You can merge it now the close the issue #180 . Thanks again.

Issue #162, opened 2016-04-17
BatchProcessor can write points to the wrong retention policy.


larrywalker:
We are writing 2 events to 2 different tables at the same time each with their own retention policy (RP)

With batching enabled, if you are writing to measurements with different RP, one of the measurements can sometimes be written with the wrong RP.  Then a query of the data with the correct RP makes it seem as though the measurement did not get written.

If stepping through the code, it slows down the writes enough that the events get put into different batches and therefore they get put in the correct measurements with the correct RP.

ONLY ONE EVENT STORED (EVENTS BATCHED INTO 1 RP)
[http] 2016/04/17 05:58:58 127.0.0.1 - root [17/Apr/2016:05:58:58 -0400] GET /query?db=dev&epoch=ms&p=%5BREDACTED%5D&q=SELECT+%2A+FROM+evttest_rp.evttest+WHERE+deveui%3D%2700250C0100000300%27+AND+time+%3E+now%28%29+-+86400s&u=root HTTP/1.1 200 2933 - okhttp/2.7.5 01037114-0483-11e6-910c-000000000000 11.485795ms
[wal] 2016/04/17 05:59:03 Flush due to idle. Flushing 48 series with 48 points and 1968 bytes from partition 1
[wal] 2016/04/17 05:59:03 write to index of partition 1 took 6.381651ms
[http] 2016/04/17 05:59:03 127.0.0.1 - root [17/Apr/2016:05:59:03 -0400] **POST /write?consistency=one&db=dev&p=%5BREDACTED%5D&precision=n&rp=evtsys_rp**&u=root HTTP/1.1 204 0 - okhttp/2.7.5 041a23d6-0483-11e6-910d-000000000000 1.456205ms
[wal] 2016/04/17 05:59:09 Flush due to idle. Flushing 2 series with 2 points and 290 bytes from partition 1
[wal] 2016/04/17 05:59:09 write to index of partition 1 took 1.16645ms

BOTH EVENTS STORED CORRECT – DIFFERENT RP - Stepping through code
[wal] 2016/04/17 06:03:13 write to index of partition 1 took 6.276625ms
[wal] 2016/04/17 06:03:23 Flush due to idle. Flushing 48 series with 48 points and 1968 bytes from partition 1
[wal] 2016/04/17 06:03:23 write to index of partition 1 took 6.882536ms
[...]

PR #243, opened 2016-11-14
support async write point with different rp.

jiafu1115:
same dbname with different rp should write two batchs instead of only one with same rp.

@larrywalker  @majst01  Can you help to check if this PR can fix your issue #162

### DIFF
--- a/src/main/java/org/influxdb/impl/BatchProcessor.java
+++ b/src/main/java/org/influxdb/impl/BatchProcessor.java
@@ -186,8 +186,9 @@ public class BatchProcessor {
       if (this.queue.isEmpty()) {
         return;
       }
-
-      Map<String, BatchPoints> databaseToBatchPoints = Maps.newHashMap();
+      //for batch on HTTP.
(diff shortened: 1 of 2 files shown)


src/main/java/org/influxdb/impl/BatchProcessor.java
```diff
@@ -13,6 +13,7 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import org.apache.commons.collections4.keyvalue.MultiKey;
```
majst01: I dont want to introduce more dependencies, a almost equivalent implementation would be:
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Multimap.html

jiafu1115: @majst01  I also like less dependency and less code changes. So I will change the code. If nobody compain it isn't very strict, I think it is OK. WDTY?

jiafu1115:
@majst01 done according to your advices. We can go ahead now. Thanks.

PR #244, opened 2016-11-15
remove the limit for database name: not contain -

jiafu1115:
@majst01 
refer to https://docs.influxdata.com/influxdb/v1.1/introduction/getting_started/
A fresh install of InfluxDB has no databases (apart from the system _internal), so creating one is our first task. You can create a database with the CREATE DATABASE <db-name> InfluxQL statement, where <db-name> is the name of the database you wish to create. **Names of databases can contain any unicode character as long as the string is double-quoted**. Names can also be left unquoted if they contain only ASCII letters, digits, or underscores and do not begin with a digit.

### DIFF
--- a/src/test/java/org/influxdb/InfluxDBTest.java
+++ b/src/test/java/org/influxdb/InfluxDBTest.java
@@ -308,6 +308,30 @@ public class InfluxDBTest {
 		Assert.assertTrue(result.contains(numericDbName));
 		this.influxDB.deleteDatabase(numericDbName);
 	}
+	
+    /**
+     * Test that creating database which name is empty will throw expected exception
+     */
+    @Test(expected = IllegalArgumentException.class)
+    public void testCreateEmptyNamedDatabase() {
+        String emptyName = "";
+        this.influxDB.createDatabase(emptyName);
+    }
+
+    /**
+     * Test that creating database which name contains -
+     */
+    @Test()
+    public void testCreateDatabaseWithNameContainHyphen() {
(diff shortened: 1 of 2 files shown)


majst01:
Ok, but then we need a additional test that database creation with a name containing a "-" still works.
I know this is a leftover back from 0.8 times .

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -318,6 +318,21 @@ public void testCreateEmptyNamedDatabase() {
...
     }
 
+    /**
+     * Test that creating database which name contains -
+     */
+    @Test()
+    public void testCreateDatabaseWithNameContainHyphen() {
+        String databaseName = "123-456";
+        this.influxDB.createDatabase(databaseName);
+        try{
```
majst01: missing space

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -318,6 +318,21 @@ public void testCreateEmptyNamedDatabase() {
...
+            List<String> result = this.influxDB.describeDatabases();
+            Assert.assertTrue(result.contains(databaseName));
+        }finally{
```
majst01: same here

jiafu1115:
@majst01 done with new test and miss space fix (sorry for my ide's style, I often forget)
BTW, can you add me as one collaborator to help you maintain this project together? Thanks.

PR #246, opened 2016-11-19
junit is only used in test code

signed:
currently depending on influx-java results in junit being on the compile scope of my project.


majst01:
Good catch, thanks!

PR #249, opened 2016-11-22
fix possible over size exception when batch udp

jiafu1115:
Motivation:

UDP packge size should <=64K, So if we enable batch for write points by UDP. it may involved the IO exception when we batch too many points:
 
`java.io.IOException: Message too long
` 
 
Modifications:

   write directly for UDP in batch processor without batch.

Result:

1. avoid over size exception. If the exception occur. use should limit point's content
2. downgrade performance due to no batch for UDP now. but it is not key point due to it is in async mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,12 +80,13 @@ influxdb-java client doesn't enable gzip compress for http request body by defau
 
 UDP's support:
 
-influxdb-java client support udp protocol now. you can call followed methods directly to wirte through UDP.
+influxdb-java client support udp protocol now. you can call followed methods directly to write through UDP.
 ```
 public void write(final int udpPort, final String records);
 public void write(final int udpPort, final List<String> records);
 public void write(final int udpPort, final Point point);
 ```
+note: make sure write content's total size should not > UDP protocol's limit(64K), or you should use http instead of udp.
 
 ### Changes in 2.4
 influxdb-java now uses okhttp3 and retrofit2.  As a result, you can now pass an ``OkHttpClient.Builder``
--- a/src/main/java/org/influxdb/impl/BatchProcessor.java
+++ b/src/main/java/org/influxdb/impl/BatchProcessor.java
@@ -221,7 +221,9 @@ public class BatchProcessor {
           BatchProcessor.this.influxDB.write(batchPoints);
       }
       for (Entry<Integer, List<String>> entry : udpPortToBatchPoints.entrySet()) {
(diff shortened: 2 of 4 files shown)


jiafu1115:
This is the fastest solution to fix the issue and we can consider to improve UDP batch by partition later if needed?

majst01:
As i already stated in https://github.com/influxdata/influxdb-java/pull/239#issuecomment-259619964 im not sure why we need udp batch processing at all.

I wanted to show me how this change will improve write performance, this was not done.

For this particular issue we need to create:

1. a bunch of performance comparisons between udp single write vs. batched write performance, written as unit tests.
1. A unit tests which shows that oversized udp packets lead to MessageSizeExceptions

If this is done, we will consider how to proceed, either:
- improve batch processing to cover different Message Size situation
- skip batch processing for udp, and throw a IllegalArgument if someone will try to do so.

jiafu1115:
@majst01 write two points into one data package should have better performance then write two points with two package. The current issue is that we can't put too many points into one package due to UDP limit.
Whatever, I will add unit test for oversized or performance compare. 
in short, batch is better, thus, we cann't batch too many due to UDP limit, That is the key point.

majst01:
@jiafu1115 "should" is short for i guess, or i dont know. Only seeing is believing/knowing

I will not accept any pr´s on this topic without the 2 questions from above answered.

jiafu1115:
@majst01  I will prove it due to I should take responsibility for my code! Thanks

jiafu1115:
@majst01  check new add test for performance compare and exception cases:

for performance: 

```
performance(ms):write udp with batch of 1000 string:29
performance(ms):write udp with 1000 single strings:173
```

for exception over udp limit:
I also add one test case;

for async mode udp, we should handle to avoid this exception: 
for sync mode udp, we had no need to handle it. The user should know if using udp, there is limit for size.
what's more, I had added into readme.

WDTY?  @majst01

majst01:
Hi @jiafu1115 now i had time to read all aspects and my conclusion is like yours, disable batching for UDP ist the best option. To solve this in a correct way we would need to figure out what exact environment is in place like MTU and SO_SNDBUF and then resend packets on errors etc. This is not the way to go in our library.

I merge this right now. This should be the last PR for 2.5.
What is missing from my perspective is a valid CHANGELOG.md. Are you willing to collect some points, i will do as well.

jiafu1115:
ok.Thanks. I will collect all points for changing log from my view and co-work with you to prepare changes in readme.

PR #251, opened 2016-11-30
added new method tag(Map<String, String> tags) to BatchPoints builder

christiangda:
### List of chages
* Added new method tag(Map<String, String> tags) to BatchPoints builder

### DIFF
--- a/src/main/java/org/influxdb/dto/BatchPoints.java
+++ b/src/main/java/org/influxdb/dto/BatchPoints.java
@@ -84,6 +84,18 @@ public class BatchPoints {
     }
 
     /**
+     * Add a Collection tags to this set of points.
+     *
+     * @param tags
+     *            Collections of tags
+     * @return the Builder instance.
+     */
+    public Builder tag(Map<String, String> tags) {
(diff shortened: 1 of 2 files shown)


majst01:
Hi,

to add new functionality you need to:
1. Create an issue which describes the problem statement
1. create a small PR which includes a description how implementation was made, the implementation with correct javadoc, the unit-tests with meaningful method names, additional documentation to README.md.

Looking to this i found on first sight:
Point.Builder.tag(Map<String,String> tags) is already present. You added it to Batchpoints and your implemented test is against existing code ?

So no this is not a valid PR.

majst01:
ok, but your actual approach was simply wrong.

christiangda:
Ohh sorry for confusing you, I didn't try to implement this pattern in your code,  for my  case it will be easy with the new constructor that I implement, .tag(Map<String, String> tags), is not the same, but it work for me.


In my case I need to pass a `Map` with my tags to some method that build the `BatchPoints Builder`, so, I don't have a way to create the `BatchPointsbuilder` using my `Map`

PR #259, opened 2016-12-21
InfluxDB-Java Chunking

andyflury:
Adds support for Chunking, see https://docs.influxdata.com/influxdb/v1.1/guides/querying_data/#chunking

If chunking is enabled multiple QueryResults are delivered through a streaming connection. 

I added a new query method to InfluxDBService that takes a consumer that will process received QueryResults. An example call would look like this.

`        influxDB.query(query, 20, queryResult -> System.out.println(queryResult));`

Would you consider integration this code into master?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
  - Support writing by UDP protocal.
  - Support gzip compress for http request body.
  - Support setting thread factory for batch processor.
+ - Support chunking
 
 #### Fixes
 
--- a/README.md
+++ b/README.md
@@ -96,6 +96,16 @@ public void write(final int udpPort, final Point point);
 ```
 note: make sure write content's total size should not > UDP protocol's limit(64K), or you should use http instead of udp.
 
+
+#### chunking support (version 2.5+ required):
+
+influxdb-java client now supports influxdb chunking. The following example uses a chunkSize of 20 and invokes the specified Consumer (e.g. System.out.println) for each received QueryResult
+```
+Query query = new Query("SELECT idle FROM cpu", dbName);
+influxDB.query(query, 20, queryResult -> System.out.println(queryResult));
+```
+
+
 ### Other Usages:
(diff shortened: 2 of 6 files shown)


majst01:
Looks good but code-coverage decreased which is not good. Can you please add some more corner cases to you unit-test ?

majst01:
OK, Tests are failing because we support influxdb since 0.13, and 0.13 and 1.0 dont have chunking available. So the call to chunked query must result in a proper exception to the caller.

andyflury:
Not sure how to cover those remaining edge cases. 

- The onFailure (line 377) is "invoked when a network exception occurred talking to the server or when an unexpected exception occurred creating the request or processing the response".  And I have no idea how to force that in CI. 
- And fromJson throws and IOException (line 371) in case the body cannot be parsed into a QueryResult. And again I cannot think of any way how to force that

majst01:
Ok, i will take that, can you before i merge add a changelog entry and a Example howto use this feature.

Thanks for contributing

andyflury:
I have no idea why tests are failing now, they passed before and I only updated the CHANGELOG and README. All tests pass locally using 10.13, 1.0 and 1.0

majst01:
Stacktrace is weird, you need to add some logic to unit-test to nail down this problem. I cant merge if we have failing tests.

andyflury:
ok, got it, need to make sure the DB did not get deleted before reading

majst01:
Fine, thanks for your contribution !

joelmarty:
There is an "issue" with this, the chunking implementation uses java 8 features (java.util.function.Consumer) but the pom.xml and the readme were not updated to reflect this change.

Since this feature breaks the java 7 support, I would expect some sort of public notice.

majst01:
Huh, this was my fault, i didnt notice. I thought id wouldnt compile if some Java8 feature are used because compiler is set to Java7 as well in the pom.xml even for tests.

So what did i miss ?

andyflury:
To keep compatibility with Java 7 we could substitute the java.util.function.Consumer with a new Interface e.g. org.influxdb.dto.QueryResultConsumer

PR #264, opened 2017-01-03
Resolves #263 Add databaseExists method to InfluxDB interface

### LINKED ISSUE #263: Add databaseExists method to InfluxDB interface
This is the same issue described here #173 

I'll provide a PR for this feature.

oscerd:


majst01:
Would you please add a Changelog entry, other than that i will merge.

oscerd:
@majst01 done, I guess it should be ok now

PR #270, opened 2017-01-13
signal end-of-chunking

andyflury:
When there are no more QueryResults coming in through chunking we should signal this to the consumer in some way. I suggest that we use some sort of "poison pill" for this. I added a special error message (with text "DONE") to the final QueryResult. I'm open to different options if someone has a better idea.

Thanks!

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -367,7 +367,9 @@ public class InfluxDBImpl implements InfluxDB {
                         throw new RuntimeException(errorBody.string());
                     }
                 } catch (EOFException e) {
-                    // do nothing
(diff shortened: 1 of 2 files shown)


andyflury:
Hello guys, did you get a chance to look at this one? I believe both changes are necessary. Open to suggestions if someone has a better solution. Thanks!

tactevo:
+1

majst01:
Hi,

I would take the first LinkedBlockingQueue enhancement, but you must add some unit tests to cover the additional code. Your second modification is a different topic and should therefore be discussed in a different PR.

andyflury:
Not sure why codecoverage fails, I added test cases for the additional lines. Can CodeCov tests be restartet?

andyflury:
Great! Thanks Stefan!

PR #271, opened 2017-01-16
Fixes #266 BatchPoints and Point should implement equals() and hashCo…

### LINKED ISSUE #266: BatchPoints and Point should implement equals() and hashCode()
In order to test that my programs correctly build the BatchPoints and Point objects, I like to use hamcrest matchers. Since the dto objects don't implement equals and hashCode, I cannot do that:

`assertThat(actualBatch, is(expectedBatch))` will always return false since we fall back to Object.equals() behavior.

Very simple :)

joelmarty:
…de()

Added equals and hashCode in Point and BatchPoints + tests


majst01:
@joelmarty Thanks for that, happy to merge it

joelmarty:
you're welcome!

PR #273, opened 2017-01-23
Use Charset name instead of relying on toString()

### LINKED ISSUE #272: IllegalCharsetNameException when connecting to InfluxDB
Hi,

I'm trying to open a connection to InfluxDB with library version 2.5:

``` java
InfluxDBFactory.connect("https://xxx.com/influx", "user", "pass");
```

But this crashes with an exception about an invalid charset name:

```
java.nio.charset.IllegalCharsetNameException: java.nio.charset.CharsetICU[UTF-8]
                      at java.nio.charset.Charset.checkCharsetName(Charset.java:201)
                      at java.nio.charset.Charset.forName(Charset.java:295)
                      at java.net.URLEncoder.encode(URLEncoder.java:57)
                      at org.influxdb.dto.Query.encode(Query.java:107)
                      at org.influxdb.impl.InfluxDBImpl.<clinit>(InfluxDBImpl.java:53)
                      at org.influxdb.InfluxDBFactory.connect(InfluxDBFactory.java:62) 
```

I'm using the library in Android version 6.0.1, Java version 1.7.

mnuessler:
Use the `Charset.name()` method to get the charset code instead of
relying on the `toString()` method of the Charset implementation. It
seems that on Android the `toString()` method does not always return the
charset code as expected.

Fixes #272.

### DIFF
--- a/src/main/java/org/influxdb/dto/Query.java
+++ b/src/main/java/org/influxdb/dto/Query.java
@@ -1,9 +1,8 @@
 package org.influxdb.dto;
 
-import com.google.common.base.Charsets;
-
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Represents a Query against Influxdb.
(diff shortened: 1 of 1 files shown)


pimeys:
Fixes the bug. Tested on Android 6.0.1.

PR #278, opened 2017-02-02
Magnet fixes

rogierslag:
We encountered some problems with the API.

1. Since some of our keys are based on user input, we noticed the api did not support those if there were backslashes in there
1. Sometimes tag values were empty, which caused influx itself to decline the data with an error

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -28,7 +28,10 @@ public class Point {
   private TimeUnit precision = TimeUnit.NANOSECONDS;
   private Map<String, Object> fields;
 
-  private static final Escaper FIELD_ESCAPER = Escapers.builder().addEscape('"', "\\\"").build();
+  private static final Escaper FIELD_ESCAPER = Escapers.builder()
+                                                      .addEscape('\\', "\\\\")
+                                                      .addEscape('"', "\\\"")
+                                                      .build();
   private static final Escaper KEY_ESCAPER = Escapers.builder()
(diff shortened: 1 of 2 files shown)


rogierslag:
Also closes https://github.com/influxdata/influxdb-java/issues/269

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -83,7 +86,9 @@ public static Builder measurement(final String measurement) {
     public Builder tag(final String tagName, final String value) {
       Preconditions.checkArgument(tagName != null);
       Preconditions.checkArgument(value != null);
-      tags.put(tagName, value);
+      if ( !tagName.isEmpty() && !value.isEmpty()) {
```
majst01: The first space after ( is breaking the build because this is not conform with the coding style.

rogierslag:
Fixed code and added tests

src/test/java/org/influxdb/dto/PointTest.java
```diff
@@ -220,6 +224,12 @@ public void testNullValueThrowsExceptionViaAddField() {
 		Point.measurement("dontcare").addField("field", (String) null);
 	}
 
+	@Test
+	public void testEmptyValuesAreIgnored() {
+		Point point = Point.measurement("dontcare").tag("key","").addField("dontcare", true).build();
```
majst01: Please add a tag with empty key as well.

rogierslag: Done :)

PR #281, opened 2017-02-08
Added unit test for issue ticket 256

ghmegan:
Two tests added. One will fail on numeric conversion of long number field. The other fails when retrieving timestamp in nanoseconds versus string format.

### DIFF
--- a/src/test/java/org/influxdb/TicketTest.java
+++ b/src/test/java/org/influxdb/TicketTest.java
@@ -25,7 +25,7 @@ import org.junit.Test;
  * @author stefan.majer [at] gmail.com
  *
  */
-public class TicketTests {
+public class TicketTest {
 
 	private InfluxDB influxDB;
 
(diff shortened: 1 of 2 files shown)


majst01:
But all checks passed ?

ghmegan:
Bah, had thrown exceptions instead of fail statements. Fixed it. Or rather, broke it.

ghmegan:
This is strange. I am getting failures when I run these tests in my local environment...

majst01:
do you use the same script which is used in the travis.yaml to run the tests ?
If not what is your environment like (OS, JDK ...)

ghmegan:
Ok. Got it sorted out. TicketTests.java wasn't being detected as a unit test file by maven. Had to rename it to TicketTest.java to get the tests to run.

majst01:
Huh, good catch, i would like to have this issue seperated. I will rename the class to the correct name, then your test is just a small one.

ghmegan:
Ok, I made a new branch with a cleaner commit history since I had already done the commit for file renaming in this branch. I am closing this pull request and opening a new pull request for the clean branch.

PR #282, opened 2017-02-09
Initial capacity on LinkedBlockingQueue inside BatchProcessor

andyflury:
The LinkedBlockingQueue inside BatchProcessor should have an initial capacity. Otherwise we might run into an OutOfMemoryError if the producer is match faster than InfluxDB.


majst01:
Thanks for you contribution.

andyflury:
Thanks Stefan, would you mind merging #270 as well. I did add additional test cases, so all new code is covered by tests. However it looks like the CodeCov report is still based on the original code. Not sure how to re-run it. Thanks

PR #284, opened 2017-02-14
Improved end of stream support for chunked responses

pjmyburg:
I've replaced the Java 8 Consumer with a custom interface that allows for a cleaner end of stream notification as opposed to creating a fake error QueryResult.
This also removes the Java 8 dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,16 @@ note: make sure write content's total size should not > UDP protocol's limit(64K
 influxdb-java client now supports influxdb chunking. The following example uses a chunkSize of 20 and invokes the specified Consumer (e.g. System.out.println) for each received QueryResult
 ```
 Query query = new Query("SELECT idle FROM cpu", dbName);
-influxDB.query(query, 20, queryResult -> System.out.println(queryResult));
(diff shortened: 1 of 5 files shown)


pjmyburg:
Not sure why the CI build fails?
`compile-and-test.sh` is passing locally and the in the Travis CI log I see everything passed against InfluxDB 1.2.0 and was busy passing against InfluxDB 1.1.1 when the test was terminated.
I've already retried it once and that didn't help.

majst01:
Same to me, this happens since a few days and i cant see any reason ?

majst01:
Closed https://github.com/influxdata/influxdb-java/commit/b700a3cce247fc26973b8d1bcb96561db40509a2

pjmyburg:
@majst01, this PR shouldn't be closed as a result of commit b700a3c.  It wasn't about the Java version.

This PR provided better end of stream support for chunked responses.  I don't think the current QueryResult with error "DONE" is a great idea.

majst01:
Cant be merged because of conflicts, there are also missing tests, i close this for now, please reopen any time if needed.

PR #287, opened 2017-02-16
Add InfluxDB.tryWrite(...) methods

thomaslee:
Recently a change was introduced to optionally bound the actions queue in `BatchProcessor`. That's a great improvement, but the blocking semantics of `queue.put(...)` aren't always desirable.

We (@neeveresearch) wanted to attempt to enqueue a point in the batch. If the queue is full, we then wanted to simply drop the data and continue on with other more important things. Due to the blocking semantics, we ended up having to roll our own batching.

This PR introduces `InfluxDB.tryWrite(...)` methods that have similar semantics to `BlockingQueue.offer(...)` when batching is enabled: if the BatchProcessor's action queue is full, return false and leave it to the caller to determine what to do next. If batching is _not_ enabled, it's equivalent to the semantics of a normal write (albeit we return `true`).

This change should be fully backward compatible with earlier releases at the source & bytecode level. I also modified the BatchProcessorTest to clean up threads that might otherwise leak.

### DIFF
--- a/src/main/java/org/influxdb/InfluxDB.java
+++ b/src/main/java/org/influxdb/InfluxDB.java
@@ -155,6 +155,18 @@ public interface InfluxDB {
   public void write(final String database, final String retentionPolicy, final Point point);
 
   /**
+   * Try to write a single Point to the database. Useful if batching is enabled with a bounded
+   * <code>actions</code> size: once the maximum batch size is reached, calls to
+   * {@link #tryWrite(String, String, Point)} will return <code>false</code> until the batch is
+   * written to Influx.
+   *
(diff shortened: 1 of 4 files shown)


thomaslee:
As an aside: I had to modify maven-compiler-plugin in the POM to build against Java 1.8 due to the use of `java.util.function.Consumer` in `InfluxDBImpl`. I changed the POM back before opening up this PR, but seems like it wouldn't compile with the settings in the POM? Am I doing something silly?

majst01:
Can you please check if your requirements are solved with: https://github.com/influxdata/influxdb-java/pull/319, if yes please close this PR.

majst01:
Im closing this because there was no response for a long time, @thomaslee please reopen if needed.

Issue #289, opened 2017-02-25
Batching Enhancements


jganoff:
I'm building an InfluxDB producer that I'd like to guarantee at-least-once delivery for. My throughput requirements are high enough that batching is required. I was happy when I found BatchProcessor but then realized a couple of shortcomings that are preventing me from using it. I'd rather improve this library than create my own batch producer. Here's where the current batch implementation is falling short. Are any of these being worked on currently?

1. There's no way to request the batch buffer to be flushed and not destroy the backing thread pool. The only option I see possible today without resorting to reflection is calling `InfluxDB.disableBatch` followed by `InfluxDB.enableBatch` to force a flush but that creates a new executor every time.
1. `BatchProcessor.write()` is not thread safe and may be called concurrently from `BatchProcessor.flush()` on the current thread and the scheduled timer on the thread from the pool. This results in duplicate records being sent during `flush()`.
1. Batching records can silently fail. I'd like to provide a callback with each asynchronous write request that would be invoked once the point is sent or if an error was encountered while attempting to send it.
1. The consistency level for all BatchPoints produced by the BatchProcessor is `ConsistencyLevel.ONE`. Direct control over the consistently level would be nice but ONE is a show stopper.

Some of the above are fairly substantial deviations in behavior from the current version. Would it be worthwhile for this to be a separate batch implementation rather than migrating the current version? Perhaps create a new asynchronous interface named `InfluxDBAsync`?

jganoff:
My apologies, I was wrong about 2 above. No duplicate records are written during `flush()` since `BlockingQueue.drainTo()` will prevent concurrent drains.

majst01:
1. is easy to fix, we should introduce a InflusDB,flush() which does this. 
3. is more difficult, see https://github.com/influxdata/influxdb-java/issues/148 and some other issues as well.
4. Seems easy as well, but only on a BatchProcessor instance, not per Point.

Introduction of a new Interface is not necessary in my opinion, to achieve async write some methods must be implemented with an async signature.

Do you mind implementing 1 and 3 ?

jganoff:
Sure, I'll send a PR for 1 and 3. Unfortunately I won't be able to use the built in batch support until it's at least possible to be alerted when records are not delivered. Not sure how I missed #148 but that's a good list of enhancements. Hopefully a combination of those will improve the batch implementation enough that I can use it. If not I'll see what the gap is and propose additional work.

[...]

PR #290, opened 2017-02-27
Added InfluxDB.flush() and consistency level can be configured for batching

jganoff:
This addresses a couple requested enhancements described in #289. Specifically:

1. `InfluxDB.flush()` was introduced as a mechanism for flushing the underlying `BatchProcessor` without shutting it down. It's a step toward being able to guarantee all asynchronously queued points are delivered at a given time when operating in batch mode. This doesn't change the failure handling for `BatchProcessor.write()` so records may still be silently dropped as a result of flushing.
2. The consistently level can be configured for asynchronous writes. The default was not changed and is still `ONE`. It can be explicitly set via `InfluxDB.enableBatch(int, int, TimeUnit, ConsistencyLevel)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
  - Support chunking
  - Add a databaseExists method to InfluxDB interface
+ - [Issue #289] (https://github.com/influxdata/influxdb-java/issues/289) Batching enhancements: Pending asynchronous writes can be explicitly flushed via `InfluxDB.flush()`.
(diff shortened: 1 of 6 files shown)


majst01:
Hi @jganoff 

this is a good start, but i want to seperate these 2 modification into 2 seperate PR's to make them easier to review and accept.

jganoff:
Sure, I'll split them and send an update.

PR #291, opened 2017-02-27
Allow Point.Builder to be reused for unique value, but shared tags

bentatham:
I have a case where I want to build a whole bunch of points (from existing data in other formats), and thus I want to use the same tags over and over.  I tried to do this with the ``Point.Builder``, but discovered that it passes the mutable maps of fields and tags directly to the ``Point`` it builds.

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -208,7 +208,7 @@ public class Point {
           .checkArgument(this.fields.size() > 0,
           "Point must have at least one field specified.");
       Point point = new Point();
-      point.setFields(this.fields);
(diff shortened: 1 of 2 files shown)


bentatham:
I would understand if this was a bit contentious because it causes a copy of the maps in what is likely the normal case so no one else seems to have run into this issue.  The alternative to this change would be to throw an ``IllegalStateException`` if you try to modify the ``Builder`` after calling ``build()``.  This would alert/prevent future users of the library from making the same mistake I did.

majst01:
Hi @bentatham 
sorry i dont understand neither your point, nor your implementation, can you please show a more descriptive example to understand what you want to achieve.

PR #296, opened 2017-03-06
Improve performance of TimeUtil

tdeleeuw:
TimeUtil creates 2 SimpleDateFormatter's (one for the date part and one for the time part) for each operation (toInfluxDBTimeFormat and fromInfluxDBTimeFormat).
This is suboptimal as creating a SimpleDateFormat is expensive (but ensures thread safety) and also force 2 operations.

Changing the SimpleDateFormat format to "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'" allows that only one SimpleDateFormatter can be used to parse/format the influxDBTimeFormat.

In order to reduce the number of instanciations while making sure it remains thread safe, I have introduced a ThreadLocale storing one SimpleDateFormatter object per thread. So we can use those objects without any synchronization.

### DIFF
--- a/src/main/java/org/influxdb/impl/TimeUtil.java
+++ b/src/main/java/org/influxdb/impl/TimeUtil.java
@@ -12,8 +12,17 @@ import java.util.concurrent.TimeUnit;
  *
  */
 public enum TimeUtil {
-  INSTANCE;
+    INSTANCE;
 
+    private static final ThreadLocal<SimpleDateFormat> formatter = new ThreadLocal<SimpleDateFormat>(){
+        @Override
+        protected SimpleDateFormat initialValue()
+        {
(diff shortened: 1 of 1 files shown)


majst01:
Please add some tests to show that behavior didn't change.

tdeleeuw:
Hi
I relied on TimeUtilTest which already do some sanity checks - obviously if the patch would have changed the behaviour, those tests (even if very basic) would have failed.
Could you clarify which extra tests you would like to have?
Thanks
Thierry

majst01:
You are rigtht.

tdeleeuw:
Actually I see that this code is not called from anywhere outside the unit test ;-(
Wouldn't it be better to remove it then or you prefer to keep it for backward compatibility?

PR #297, opened 2017-03-08
Add support for seconds only formatted time.

tdeleeuw:
While parsing query results I received a time that was formatted like '2017-02-05T11:00:00Z'. As you can see no millis where provided, the query was something like 
    SELECT time, data FROM myTable WHERE time >= '2017-01-01T00:00:00.000Z'
I used the Query object and retrieved the Series object
 QueryResult qr = db.query(q);
        for (QueryResult.Result r : qr.getResults()) {
            for (QueryResult.Series s : r.getSeries()) {
                for (List<Object> v : s.getValues()) {
                   //...
                  String myTimestamp = (String) v.get(timePos);

So I adapted the TimeUtil to support this format by detecting that the length is 20 insteead of 24 (as the ".xyz" is not provided)

### DIFF
--- a/src/main/java/org/influxdb/impl/TimeUtil.java
+++ b/src/main/java/org/influxdb/impl/TimeUtil.java
@@ -13,10 +13,19 @@ import java.util.concurrent.TimeUnit;
 public enum TimeUtil {
     INSTANCE;
 
-    private static final ThreadLocal<SimpleDateFormat> FORMATTER = new ThreadLocal<SimpleDateFormat>() {
+    private static final ThreadLocal<SimpleDateFormat> FORMATTER_MILLIS = new ThreadLocal<SimpleDateFormat>() {
         @Override
         protected SimpleDateFormat initialValue() {
             SimpleDateFormat dateDF = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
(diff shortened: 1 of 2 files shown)


tdeleeuw:
Checkstyle rule "20 is a magic number" is breaking the build.
I have multiple options:
1. //CHECKSTYLE:OFF
2. time.length() == (354 - 334) or (time.length() == "2017-02-05T11:00:00Z".length())
3. we disable this "rule" from checkstyle

I personally vote for 3 ;-)

please let me know how to proceed

src/main/java/org/influxdb/impl/TimeUtil.java
```diff
@@ -65,19 +75,22 @@ public static String toTimePrecision(final TimeUnit t) {
...
-     * influxdb time format example: 2016-10-31T06:52:20.020Z
+     * influxdb time format example: 2016-10-31T06:52:20.020Z or 2016-10-31T06:52:20Z
      *
      * @param time timestamp to use, in influxdb datetime format
      * @return time in unix epoch time
      */
     public static long fromInfluxDBTimeFormat(final String time) {
         try {
-            return FORMATTER.get().parse(time).getTime();
+            if (time.length() == 20) {
```
majst01: You have to introduce a final static variable for this purpose.

PR #305, opened 2017-03-29
Shading google guava to avoid dependency problems

fernandopeinado:
Like they did in elastic search and many other libraries, I've shaded google guava dependency, because I have conflicts of guava's version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,6 +111,48 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>3.0.0</version>
+				<executions>
+					<execution>
+						<phase>package</phase>
+						<goals>
+							<goal>shade</goal>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -12,7 +12,7 @@
 	<groupId>org.influxdb</groupId>
 	<artifactId>influxdb-java</artifactId>
 	<packaging>jar</packaging>
-	<version>2.5-SNAPSHOT</version>
+	<version>2.6-SNAPSHOT</version>
```
majst01: Please do not increment version, otherwise looks good

pom.xml
```diff
@@ -111,6 +111,48 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>2.1</version>
```
majst01: Actual version of maven shade plugin is 3.0

majst01: If modify as requested i will merge

PR #310, opened 2017-04-13
Delete measurement by tags, drop measurement

andreevym:
add methods in API InfluxDB: 'dropMeasurement' and 'delete'
add check method: testDelete, testDropMeasurement

### DIFF
--- a/src/main/java/org/influxdb/InfluxDB.java
+++ b/src/main/java/org/influxdb/InfluxDB.java
@@ -10,6 +10,7 @@ import org.influxdb.dto.Point;
 import org.influxdb.dto.Pong;
 import org.influxdb.dto.Query;
 import org.influxdb.dto.QueryResult;
+import org.influxdb.exception.DeleteInfluxException;
 
 /**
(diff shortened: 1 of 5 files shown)


majst01:
This needs a rebase, not mergable in the current state, i am also not sure if we should make the api wider ? Could we move this kind of functionality into a seperate api ? What do you guys think ?

lxhoan:
@ivankudibal and I recently discussed that seems this PR is not too necessary

I agree that we should keep the client API thin and consistent to InfluxDB REST API (https://docs.influxdata.com/influxdb/v1.5/tools/api/, where only 'write' and 'query' available for CRUD)

BTW, in my opinion, keeping the client API thin and consistent to InfluxDB REST API can help to reach a common guideline for designing/developing new client API in other programming language/environment

So I propose to close this PR

andreevym:
my team for use "DELETE" is normal and useful, but I got it
If you want, you can move this in utils

lxhoan:
Hi @majst01 , so that means we can close this PR ?

PR #311, opened 2017-04-16
Add a query() method that returns an Iterator<QueryResult>

Dromit:
New query() methods as mentioned in https://github.com/influxdata/influxdb-java/issues/308

```java
  public Iterator<QueryResult> query(Query query, int chunkSize);
  public Iterator<QueryResult> query(final Query query, int chunkSize, TimeUnit timeUnit);
```

### DIFF
--- a/src/main/java/org/influxdb/InfluxDB.java
+++ b/src/main/java/org/influxdb/InfluxDB.java
@@ -1,5 +1,6 @@
 package org.influxdb;
 
+import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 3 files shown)


theoDiefenthal:
For the Iterator<QueryResult>, you make a `call.execute()`, which is synchronous. I didn't test the code but doesn't it mean that the whole result is obtained first and then you can iterate over it one by one? The point of chunking is do avoid this explicitly by being able to fetch one result after another while influx  is still sending...

majst01:
no test, no response to questions, closing. Please reopen any time if required.

PR #312, opened 2017-04-19
Update Java Requirement

jackzampolin:
https://community.influxdata.com/t/influxdb-java-api-not-for-jdk-1-7-but-1-8/635


majst01:
This PR https://github.com/influxdata/influxdb-java/pull/284 does the opposite. I am keen to switch to java with version 2.6, so i will modify the pom.xml and the readme an the close this one.

Issue #275, opened 2017-01-24
Add return information for write


lionelmazeyrat:
When we use a write operation like : influxDB.write(batchPoints);
It will be very usefull to have a return object with the list of status (good/bad) for each write points operation.

skovzhaw:
Must have feature for us too...

majst01:
I am happy to review PR's

yuguangchuan:
So why influxdb's java client doesn't provide the function of returning a status code for `influxDB.write(batchPoints)`? And in current cases, how does client know whether the batchPoints is successfully written or not?

majst01:
What should the status code look like ?
- Number of Points written ?

To have this information the write method must wait for all batchpoints to be written and will therefore be synchronous which is a disaster from performance perspective.

Again if someone is willing to provide a PR im happy to review.

lionelmazeyrat:
To avoid to reduce write performance, perhaps the write method could implement an interface to be asynchronously notified of the write completion with a list of status for each points ?

yuguangchuan:
i think the the status code should be provided in `write(batchpoints)` method, such as whether the batch was successfully written or not instead of "_Number of Points written_". Thus the client could decide if it needs to  retry/rewrite this batch or not. Note that InfluxDB's ability to write points by batch is strong(See [Influxdb - Batch](https://docs.influxdata.com/influxdb/v1.2/concepts/glossary/#batch)), so it won't be considered as a disaster in synchronous write pattern.

From higher view, this influxdb-java API provides two ways of points writing: `write(batchpoints)` and `enablebatch`, but they didn't show significant difference. It will be better that `write(batchpoints)` should be designed as a synchronous method for those clients who need to ensure data accuracy and `enablebatch` should be designed for those who ask for high write performance with tolerance of data loss. I think this API will look much better in this perspective.

Just some personal idea :-)

fmachado:
**Just for reference**

HBase API allows you to batch async actions[1] and, by using an array passed as parameter to the batch call, get or inspect the result of every async call. As you can see here[2], every async call may result with:

- a _null_ value (if the action is not supposed to return anything);
- an instance of _Throwable_;
- an instance of org.apache.hadoop.hbase.client.AsyncProcess.ReplicaResultState (if the async call is still in progress);
- a Java wrapper class or String (e.g. the value stored in HBase).

[...]

PR #318, opened 2017-05-04
Add exception callback to handle async errors

csokol:
This enables the client to do something (other than logging) if a problem happens asynchronously. In our use case, we want to send the error to honeybadger.io to be notified of influxdb issues.

### DIFF
--- a/src/test/java/org/influxdb/impl/BatchProcessorTest.java
+++ b/src/test/java/org/influxdb/impl/BatchProcessorTest.java
@@ -9,6 +9,7 @@ import static org.mockito.Mockito.verifyNoMoreInteractions;
 
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 
 import org.influxdb.InfluxDB;
 import org.influxdb.dto.BatchPoints;
@@ -40,6 +41,25 @@ public class BatchProcessorTest {
         // without try catch the 2nd time does not occur
         verify(mockInfluxDB, times(2)).write(any(BatchPoints.class));
     }
+
+  @Test
+  public void testSchedulerExceptionHandlingCallback() throws InterruptedException, IOException {
+    InfluxDB mockInfluxDB = mock(InfluxDBImpl.class);
+    Consumer<Throwable> mockHandler = mock(Consumer.class);
+    BatchProcessor batchProcessor = BatchProcessor.builder(mockInfluxDB).actions(Integer.MAX_VALUE)
+        .interval(1, TimeUnit.NANOSECONDS).exceptionHandler(mockHandler).build();
+
+    doThrow(new RuntimeException()).when(mockInfluxDB).write(any(BatchPoints.class));
+
+    Point point = Point.measurement("cpu").field("6", "").build();
+    BatchProcessor.HttpBatchEntry batchEntry1 = new BatchProcessor.HttpBatchEntry(point, "db1", "");
+    BatchProcessor.HttpBatchEntry batchEntry2 = new BatchProcessor.HttpBatchEntry(point, "db2", "");
+
+    batchProcessor.put(batchEntry1);
+    Thread.sleep(200); // wait for scheduler
+
+    verify(mockHandler, times(1)).accept(any(RuntimeException.class));
+  }
 
     @Test
     public void testBatchWriteWithDifferenctRp() throws InterruptedException, IOException {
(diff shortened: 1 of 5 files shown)


src/test/java/org/influxdb/impl/BatchProcessorTest.java
```diff
@@ -41,6 +42,25 @@ public void testSchedulerExceptionHandling() throws InterruptedException, IOExce
         verify(mockInfluxDB, times(2)).write(any(BatchPoints.class));
     }
 
+  @Test
```
csokol: it looks like the tests are not formatted with the same settings as the production code. I'm not sure if that's on purpose or accidental but I decided to use the same formatting as in the main code.

majst01:
Build is failing because of formatting.

csokol:
Fixed, @majst01

majst01:
Addresses #275

majst01:
I like your simple approach, if you mind adding some usage information's to README.md i would like to merge this.

csokol:
@majst01 cool!

Now that I'm writing the documentation, I wondering if I should add another overloaded method `enableBatch` to `InfluxDBImpl`, something like:

```java
void enableBatch(int actions, int flushDuration, 
  TimeUnit flushDurationTimeUnit, Consumer<Throwable> exceptionHandler)
```

What do you think? Is there another way to enable batching? Or maybe I should just add `void enableBatch(BatchProcessor batchProcessor)`, maybe this allows more flexibility on the caller side.

majst01:
Yes, would be a good Idea to have a second variation of enableBatch.

csokol:
which one?  `enableBatch(BatchProcessor batchProcessor)` or `enableBatch(int actions, int flushDuration, TimeUnit flushDurationTimeUnit, Consumer<Throwable> exceptionHandler)`?

I prefer the first one.

kkarski:
Is the client currently java 7 compatible? Since the Consumer<T> was introduced in java8, it would break any backwards compatibility.

majst01:
related to: #289

kkarski:
Per my similar changes referenced in #289 which I commented on, my use case requires retrying to persist the failed batch of points so the exception handling method needs to reference the failed points in the batch so they can be re-tried.

Also, flip side to that, when the batch succeeds, my use case also requires other parts of the system to be notified so I added a onSuccess event as well as part of the same interface.

Overall, our solutions are very similar.

majst01:
Having a list of failed points available in the ExceptionHandler seems a useful addition. But i would like to get the first step done, then we can add further improvements.

@csokol what do you think ?
@csokol Checkstyle is failing again

kkarski:
The only thing I'd point out then is that the Consumer functional interface chosen for this change only accepts one input, in this case the Throwable. Should you want to pass more, Consumer may not be the best choice.

csokol:
@majst01 agree. The formatting is fixed.

@kkarski passing the failed points to the callback sounds good, but I think that for your use case (retrying failed writes) it would be better to have that as a library feature. Maybe the user should be able to define a retry policy like "retry five times with exponential backoff increasing time"

kkarski:
@csokol I agree an automated retry would be a good feature and would love to see that. It doesn't solve the issue 100% however which is why I still think it would be good to have a reference to the failed points in the batch in order to attempt some sort of last ditch effort to no loose the data. Alternatively, if the Throwable was a custom type, the points could be referenced in the exception itself rather than as a second parameter.

As I said earlier however, I also have an interest in successful batches.

majst01:
Should we stick with the current approach and try to implement a automatic recovery of failed batch writes later ?

kkarski:
I think the automated recovery is optional but getting a reference to the failed points is necessary. Just my opinion.

majst01:
The introduction of a InfluxDBExceptionHandler which implements Consumer and holds a List of failed Batchpoints will do the trick, bonus we will get a fork user back. @csokol would you like to modify your PR this way ?

csokol:
The `InfluxDBExceptionHandler` would have to hold a mutable field with the list of batchpoints, is that what you mean? This doesn't look good as it might cause concurrency issues. Another way to do it is use something like `BiConsumer<List<Batchpoint>, Throwable>`. Then we can keep the implementation immutable and more thread safe. WDYT, @majst01?

majst01:
Sounds good, tbh i am still a bit oldschool and not very familiar with all the glories of the function interfaces in java8 :-)

If you mind modifying this PR, or create a separate one to see the differences in implementation.

csokol:
I'll do it in a separate branch to compare it. Looking at the code, I think I won't be able to get the points failed (I think I'll have to keep separate entries for udp/http points and keep removing them from a list or something). Is it ok if in this initial implementation I pass the whole batch to the consumer instead of just the ones that failed?

csokol:
Just noticed that #289 was actually proposing the same thing (the whole batch, not only the failed points) so I'll go for that.

majst01:
Replaced by a better implementation in #319

PR #322, opened 2017-05-23
Drop guava runtime dependency

simon04:
Some fuctionality is provided by Java 8, the remaining parts can easily be rewritten.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -254,6 +254,7 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>21.0</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.squareup.retrofit2</groupId>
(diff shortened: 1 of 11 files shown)


majst01:
In general i agree with this, but this PR has to much modifications, changes formatting of lines which are not affected, error messages are different, tests are failing.

If you are trying to improve this i can accept tis.

simon04:
I just fixed the checkstyle issues (which caused the tests to fail). There shouldn't be any unrelated formatting changes, though.

majst01:
Ok looks better, i wait til 2.6 is published to maven repo and we can then consider merging for the next release.

simon04:
Thanks for reviewing this PR and maintaining this Java project! :+1:

majst01:
@simon04 would you please resolve the merge conflict, after that i will merge.

simon04:
Done.

PR #323, opened 2017-05-25
Replace RuntimeException with InfluxDBException

### LINKED ISSUE #313: Subclass RuntimeException?
As far as I can see, the error handling in influxdb-java consists of Exceptions wrapped in a `RuntimeException`, like here: https://github.com/influxdata/influxdb-java/blob/master/src/main/java/org/influxdb/impl/InfluxDBImpl.java#L101
I presume that one reason for doing so was to make the Exception handling unchecked, which I am in principle not opposed to. However, this pattern forces me to surround calls to influxdb-java in try-catch blocks catching `RuntimeExceptions`. I just now ran into a case where I had a NullPointerException within this try-catch block that, being an instance of `RuntimeException` also gets caught. Wouldn't it be much better to create a custom exception, e.g. `InfluxException` that inherits from `RuntimeException`? This way client-side error handling gets much easier and if I am not mistaken, it would not even break existing code.

simon04:
Fixes #313.

### DIFF
--- a/src/main/java/org/influxdb/InfluxDBException.java
+++ b/src/main/java/org/influxdb/InfluxDBException.java
@@ -1,0 +1,21 @@
+package org.influxdb;
+
+/**
+ * A wrapper for various exceptions caused while interacting with InfluxDB.
+ *
+ * @author Simon Legner
+ */
+public class InfluxDBException extends RuntimeException {
+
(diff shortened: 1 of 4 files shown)


majst01:
Looks good, wait for next release with merging.

gnebehay:
nice, thanks!

PR #327, opened 2017-05-30
Simplify write() methods for use cases writing all points to same database and retention policy

andyfeller:
More often than not, applications are going to write points to a specific database and retention policy, however it's tedious passing around configuration for writes.  Instead there should be simpler write() methods for non-BatchPoints calls that uses defaults set at the client layer instead of the usage statement.


andyfeller:
Could use a little help identifying the checkstyle errors as helper script output isn't super helpful.

PR #330, opened 2017-06-01
Simple but significant performance improvements.

brettwooldridge:


majst01:
Can you show how big the performance improvements are ?

brettwooldridge:
Locally measured.

Current:
```
Single Point Write for 10000 writes of Points took:240.6 ms
5Mio points:7.300 s
```

New:
```
Single Point Write for 10000 writes of Points took:192.6 ms
5Mio points:5.306 s
```

brettwooldridge:
More difficult to measure, though measurable, and not insignificant, is the reduction of garbage.

majst01:
Great thanks, i will merge it in the next iteration.
Thanks for your contribution.

brettwooldridge:
@majst01 Thanks!  We may have more contributions in the coming weeks.

brettwooldridge:
Latest commit gets the ``maxWritePointsPerformance()`` test down to ``5Mio points: 4.781 s`` compared to ``5Mio points:7.300 s`` on master.  Also generates +800MB *less* garbage (yes, that test generates several gigabytes of garbage).

brettwooldridge:
@majst01  A few questions.  I see that v2.5 was released six months ago, and the general cadence seems to be 3-6mo between releases.  When is v2.6 slated for release?  And what is the reason this commit is scheduled for 2.7?   That would imply a wait of 6-9 months before we see this commit in a release version.

majst01:
Oh no, 2.6 should be released in the next few days if not already release (it depends on support from som influxdata guys).
After 2.6 is out i would like to release 2.7 very shortly afterwards, only with a few important fixes.

majst01:
2.6 is finally released.

brettwooldridge:
@majst01 Thanks!

majst01:
Hi Brett,

one last nit, can you please add a proper changelog entry, after that i will merge it. 
You should open a new 2.7 section therefore

PR #333, opened 2017-06-08
Enabling point builder for being used statically

andyfeller:
* Migrated Point.Builder.build() field assertion to InfluxDBImpl.write(..., Point) and BatchPoint.point(Point)
* Added helpers to tag, add fields, get fields to Point
* Added unit tests to assert new point building pattern worked the same as existing cases

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -204,9 +204,6 @@ public class Point {
       Preconditions
           .checkArgument(!Strings.isNullOrEmpty(this.measurement),
           "Point name must not be null or empty.");
-      Preconditions
-          .checkArgument(this.fields.size() > 0,
-          "Point must have at least one field specified.");
       Point point = new Point();
       point.setFields(this.fields);
       point.setMeasurement(this.measurement);
@@ -267,6 +264,11 @@ public class Point {
    */
   void setFields(final Map<String, Object> fields) {
     this.fields = fields;
+  }
+
+  public Map<String, Object> getFields() {
+
+    return fields;
   }
 
   @Override
@@ -385,4 +387,51 @@ public class Point {
     return sb;
   }
 
+  public Point tag(final String tagName, final String value) {
+
+    Preconditions.checkArgument(tagName != null);
+    Preconditions.checkArgument(value != null);
+    if (!tagName.isEmpty() && !value.isEmpty()) {
+      tags.put(tagName, value);
+    }
+    return this;
+  }
+
+  public Point tag(final Map<String, String> tagsToAdd) {
+
+    for (Entry<String, String> tag : tagsToAdd.entrySet()) {
+      tag(tag.getKey(), tag.getValue());
+    }
+
+    return this;
+  }
+
+  public Point addField(final String field, final boolean value) {
+    fields.put(field, value);
+    return this;
+  }
+
+  public Point addField(final String field, final long value) {
+    fields.put(field, value);
+    return this;
+  }
+
+  public Point addField(final String field, final double value) {
+    fields.put(field, value);
+    return this;
+  }
+
+  public Point addField(final String field, final Number value) {
+    fields.put(field, value);
+    return this;
+  }
+
+  public Point addField(final String field, final String value) {
+    if (value == null) {
+      throw new IllegalArgumentException("Field value cannot be null");
+    }
+
+    fields.put(field, value);
+    return this;
+  }
 }
--- a/src/test/java/org/influxdb/InfluxDBTest.java
+++ b/src/test/java/org/influxdb/InfluxDBTest.java
@@ -650,4 +650,20 @@ public class InfluxDBTest {
         this.influxDB.flush();
     }
 
+    @Test(expected = IllegalArgumentException.class)
+	public void testWritePointMissingFieldsUDP() {
+
+		String measurement = TestUtils.getRandomMeasurement();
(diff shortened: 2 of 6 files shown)


majst01:
Can you please explain what problems you want to solve ?

andyfeller:
The Point.Builder class is a factory for creating Point instances, which I prefer to instantiate statically and reuse throughout the life of the code.  In it's current implementation, Point.Builder.build() requires 1+ fields are set on the builder, which isn't necessarily wanted as values typically vary.

```java
class Measurements {
    public static final Point.Builder THING = Point.measurement("thing");
}

class ThingDoer {
    private final InfluxDB influxDB;

    public ThingDoer(final InfluxDB influxDB) {
        this.influxDB = influxDB;
    }

    public void doSomething(Thing thing) {

        // ...
        influxDB.write(Measurements.THING
            .build()
            .tag("class", thing.getClass().getName(),
            .field("value", 1));
    }
}
```

In the case above, I want to minimize instantiating Point.Builder class for high volume point collection as well as letting code decide what tags and fields are relevant.  I presume the logic ensuring there is 1+ fields is really only required at the time of writing points, hence moving of the logic elsewhere.  The only time I want to set tags and fields on Point.Builder instances are for things I want to capture ubiquitously, which is few and far between.

I'd welcome an advanced example of instrumenting a reasonably complex Java application with influxdb-java client including handling of wiring of InfluxDB client and other best practices for building points if this pattern seems unreasonable.  I haven't found anyone that has talked about their experiences instrumenting and what they do.

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -269,6 +266,11 @@ void setFields(final Map<String, Object> fields) {
     this.fields = fields;
   }
 
+  public Map<String, Object> getFields() {
```
majst01: This makes the Point Object mutable, i really want to prevent that.

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -385,4 +387,51 @@ private StringBuilder formatedTime() {
     return sb;
   }
 
+  public Point tag(final String tagName, final String value) {
+
```
majst01: useless newlines, not consistent with other style.

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -650,4 +650,20 @@ public void testFlushThrowsIfBatchingIsNotEnabled() {
         this.influxDB.flush();
     }
 
+    @Test(expected = IllegalArgumentException.class)
```
majst01: spaces mangled.

src/main/java/org/influxdb/dto/Point.java
andyfeller: Technically the object is mutable in the sense that other properties of the object can be changed after being built.  That aside, could you elaborate on when and why should objects be immutable?

src/main/java/org/influxdb/dto/Point.java
andyfeller: Is there a specific tool you use to format code against a predetermined style configuration?

Despite different perspectives, it's easier to conform if it's easy to do and well known.

src/main/java/org/influxdb/dto/Point.java
majst01: Immutablity the main goal of having builders, you gain guaranteed correct objects and faster synchronization. See: Josh Bloch "effective Jave" e.g.: http://www.informit.com/articles/article.aspx?p=1216151&seqNum=2

src/main/java/org/influxdb/dto/Point.java
majst01: most is specified in checkstyle.xml

src/main/java/org/influxdb/dto/Point.java
andyfeller: I was hoping you could cite actual challenges / problems with the client to date that require it to be immutable rather than point out an ideal best practice that itself cites trade offs of using such a technique:

> While the cost of creating the builder is unlikely to be noticeable in practice, it could be a problem in some performance-critical situations

I encourage re-examining this practice as the Builder pattern was designed to address issues in Factory implementations, which have changed significantly with the advent of Spring Framework and other DI frameworks.

Ultimately, Point is not immutable (setFields, setPrecision, setTags, setTime, setMeasurement, etc) so I'd think there is room for leeway.

src/main/java/org/influxdb/dto/Point.java
andyfeller: Is there a specific tool that will use checkstyle.xml to format the code according to the rules / standards defined within?

majst01: asking twice did not fix the formatting issues, without them fixed im wont merge.

andyfeller: I'm not disagreeing with the coding standard; I'm asking if there is a tool that will do the formatting according to the standard.  Seriously I'm trying to satisfy what you're asking for as I'm not familiar with tools compatible with checkstyle and whether they will fix the code in question.

majst01: not at the moment, i just defined some checkstyle rules, which are not covering all as we can see.
But i am thinking about introducing automatic formatting with for example: https://github.com/google/google-java-format

But this can only be done at the very beginning of a sprint.

majst01:
Not a single comment was accepted, closing

PR #336, opened 2017-06-12
Optimize Point.lineProtocol()

bentatham:
While testing high throughput, I profiled the sender, and discovered it was creating a new ``NumberFormat`` for every point, whether or not it actually ever needed it. I've changed it to use a ``ThreadLocal`` of it (since ``NumberFormat`` is not thread-safe), and to reuse the same instance. I've also changed it to use a single ``StringBuilder`` rather than returning strings (built by additional instances) and appending them together.


majst01:
Hi,

i think this is duplicate of #330, which will be merged soon.

bentatham:
lol -- so it is - nice to know that two of us got to some very similar optimizations independently.

PR #338, opened 2017-06-15
propagate IOExceptions to consumer

andyflury:
We should let the consumer know if there is a IOException and then he can decide what he wants to do. At the moment we simply throw the exception and since this happens in a separate thread the consumer will not be notified.

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -97,6 +97,24 @@ public class InfluxDBImpl implements InfluxDB {
     this.adapter = moshi.adapter(QueryResult.class);
   }
 
+    InfluxDBImpl(final String url, final String username, final String password, final OkHttpClient.Builder client,
+            final InfluxDBService influxDBService, final JsonAdapter<QueryResult> adapter) {
+        super();
+        this.hostAddress = parseHostAddress(url);
+        this.username = username;
+        this.password = password;
+        this.loggingInterceptor = new HttpLoggingInterceptor();
+        this.loggingInterceptor.setLevel(Level.NONE);
+        this.gzipRequestInterceptor = new GzipRequestInterceptor();
+        this.retrofit = new Retrofit.Builder()
+                .baseUrl(url)
+                .client(client.addInterceptor(loggingInterceptor).addInterceptor(gzipRequestInterceptor).build())
+                .addConverterFactory(MoshiConverterFactory.create())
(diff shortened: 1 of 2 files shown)


majst01:
This is ok, but can you add a test ?

andyflury:
I tried, problem is, I don't know how to force an IOException from the server side during testing, any ideas?

majst01:
You have to use a mock implementation

andyflury:
Can you point me to an example?

andyflury:
Hi Stefan, I'll need your help on this, can you please point me in the right direction on how to prepare a mock implementation? The mock still needs to run the InfluxDBImpl.query which in turn needs to access the private influxDBService. Thanks

andyflury:
@majst01, did you get a chance to look at this? I would need your help on the test case

majst01:
Hi @andyflury 
Im quite bussy at the moment, there are some tests already present which mock several aspects of the library with the mockito framework, probably this helps.

majst01:
Hi,

almost:
```
/home/travis/build/influxdata/influxdb-java/src/main/java/org/influxdb/impl/InfluxDBImpl.java:100: error: Line is longer than 120 characters (found 184).
/home/travis/build/influxdata/influxdb-java/src/main/java/org/influxdb/impl/InfluxDBImpl.java:100:117: error: Parameter influxDBService should be final.
/home/travis/build/influxdata/influxdb-java/src/main/java/org/influxdb/impl/InfluxDBImpl.java:100:150: error: Parameter adapter should be final.
```

andyflury:
Hi @majst01, I had to mock quite a few elements, but now I was able to test both IOException and EOFException. Can you merge it like this?

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -97,6 +97,24 @@ public InfluxDBImpl(final String url, final String username, final String passwo
     this.adapter = moshi.adapter(QueryResult.class);
   }
 
+    InfluxDBImpl(final String url, final String username, final String password, final OkHttpClient.Builder client,
```
majst01: not exactly sure why we need a new Constructor ?

andyflury: I don't see any other way how to mock the InfluxDBService and the JsonAdapter (which will need to throw the Exceptions). The constructor has package visibility so that it can only be called from the same package (i.e. the Unit Test).

Issue #340, opened 2017-06-20
QueryResult to Object mapper


fmachado:
I had some issues in the past handling the QueryResult and the internal data structure. In order to make the result manipulation more simple, I created an object mapper between the QueryResult and an annotated POJO.

https://github.com/fmachado/influxdb-pojomapper

Do you have any plans to add a similar feature to influxdb-java?

BTW, thank you Stefan and everyone here working to maintain this library.

majst01:
Hi @fmachado,

This would be a great enhancement to influxdb-java ! I would love to integrate it. If you mind you can create a pull request for the integration and we can discuss on that how to proceed.

fmachado:
@majst01 thanks for the quick response on this issue.

I'll give you a hand if there is anything reported related with this feature. Also, I've already got an OK from my employee ([azeti Networks AG](http://azeti.net)) to keep contributing with more improvements, if necessary.

I'm closing this issue.

majst01:
This is interesting, greetings to your CEO i know him personally from a former bank project in munich :-)

PR #341, opened 2017-06-21
QueryResult to Object mapper added

fmachado:
in ref to #340


majst01:
Hi @fmachado 

thanks for the effort ! 
I would prefer to have it in the same package structure not necessarily below "pojomapper".
The annotations can go into its own subpackage "annotations" but all other should be in impl.
what do you think.

fmachado:
Done. :) Do you want me to update the main README.md file?

majst01:
Excellent, a explanation which a note that this feature as since 2.7 in the Readme is needed, after this i will merge.

Issue #343, opened 2017-06-27
Is it impossible to get tag using 'InfluxDBResultMapper' with group by?


ilias-y:
Hi, InfluxDBResultMapper helps me to use influxdb-java. 
But I wonder how to use tag in '@Column(name="a", tag=true)'. I want to get the tags from QueryResult with toPOJO(). is it impossible to get the tags of series in QueryResult?

This is my QueryResult.
`QueryResult [results=[Result [series=[Series [name=tb_network, tags={CARRIER=000/00, DEVICE_OS_VERSION=4.4.2}, columns=[time, median, min, max], values=[[1970-01-01T00:00:00Z, 233.8, 0.0, 3090744.0]]], Series [name=tb_network, tags={CARRIER=000/00, DEVICE_OS_VERSION=9.3.5}, columns=[time, median, min, max], values=[[1970-01-01T00:00:00Z, 552.0, 135.0, 267705.0]]]], error=null]], error=null]`

I got the values but not the tags.

fmachado:
@ilias-y you are right: "group by" results are not currently supported but I'm working on it.

"Stay tuned." :)

ilias-y:
I'm looking forward to work it. When do you plan to complete this?

fmachado:
@ilias-y could you please test it?
Just a heads up: InfluxDB tags are always _String_ and currently we do not support Class field declarations other than... String. :)

PR #345, opened 2017-06-28
Added support map queryresult groupby

fmachado:
When the "GROUP BY" clause is used, "tags" are returned as Map<String,String>.
Accordingly with InfluxDB documentation https://docs.influxdata.com/influxdb/v1.2/concepts/glossary/#tag-value, "TAG" is always a String so currently we are not supporting (at)Column annotation on non-String Class fields.
This solves the issue #343.


majst01:
@fmachado would you mind adding a Changelog entry and extended documentation ?

btw: thanks for this.

bukajsytlos:
Hello.
We really need use this feature. 
I would like to ask when you plan to release this?

Thank you.

PR #348, opened 2017-06-30
Extend Result and Series POJO for Influx REST API sent attributes when chunking is used

theoDiefenthal:
...

### DIFF
--- a/src/main/java/org/influxdb/dto/QueryResult.java
+++ b/src/main/java/org/influxdb/dto/QueryResult.java
@@ -1,5 +1,7 @@
 package org.influxdb.dto;
 
+import com.squareup.moshi.Json;
+
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/influxdb/dto/QueryResult.java
```diff
@@ -69,6 +71,9 @@ public void setError(final String error) {
   }
 
   public static class Result {
+    @Json(name = "statement_id") private long statementId = 0;
```
fmachado: with primitives, default is `0` for integers and `false` for boolean so, just remove this please.

fmachado:
@theoDiefenthal I want to merge this PR but I cannot proceed without a proper description. In the PR title you mentioned _Extend_ but it seems you are fixing the existing objects by adding two missing properties from the JSON response when querying InfluxDB.

Could you please add an example of the original JSON with these missing properties or link to a InfluxDB documentation?

Is there a way to test this?

You added the annotation `@Json(name = "statement_id")`. Is it possible to have a JSON response without this property? If so, what happens when Moshi tries to parse it: will it throw an exception as Jackson does or it will silently ignore?

PR #350, opened 2017-07-05
Let InfluxDB handle the timestamp if none provided.

### LINKED ISSUE #349: Add possibility to let InfluxDB handle the timestamp itself
Due to a [bad support of accurate timestamps](https://stackoverflow.com/a/33487917) (especially in older Java versions) I would love to let InfluxDB handle the timestamps by itself.

> A point is uniquely identified by the measurement name, tag set, and timestamp. If you submit a new point with the same measurement, tag set, and timestamp as an existing point, the field set becomes the union of the old field set and the new field set, where any ties go to the new field set. This is the intended behavior.

[read more](https://docs.influxdata.com/influxdb/v1.2/troubleshooting/frequently-asked-questions/#how-does-influxdb-handle-duplicate-points)

As I have several data-points in a single measurement per second and it is not very recommended to have a unique-id in a tag (performance reasons) I would like to have an accurate timestamp. Otherwise I would loose data!

In the method build of Point I found the following behavior:

>             if(this.time != null) {
>                 point.setTime(this.time);
>                 point.setPrecision(this.precision);
>             } else {
>                 point.setTime(Long.valueOf(System.currentTimeMillis()));
>                 point.setPrecision(TimeUnit.MILLISECONDS);
>             }

So, if no time is provided for a certain point this library does not what I want it to do. I can't find any simple way to let InfluxDB handle the timestamp using this library?

wasnertobias:
As discussed in #349.


majst01:
thanks, can you add a Changelog entry, i will merge then

wasnertobias:
@majst01 You're welcome. Done. Hope it meets your expectations.

majst01:
Almost, 2.8 still unreleased.

wasnertobias:
@majst01 Alright, changed it. Thank you so far. :) Would be nice to see that released asap.

Issue #295, opened 2017-03-02
[feature request] Create and Drop retention policy


kostasb:
Feature request to provide a createRetentionPolicy method in a proposed form of:

`influxDB.createRetentionPolicy(rpName,dbName,duration,shardDuration,replication,default);`

Respectively, a dropRetentionPolicy method should accompany this:

`influxDB.dropRetentionPolicy(rpName,dbName);`

majst01:
If you are willing to write a PR i will review it.

PR #351, opened 2017-07-06
feature: #295 Create and Drop retention policies

gurlWhoCodes:
* Created methods to create and drop retention policies.
* Added unit test to test the same.


majst01:
Nice, would you mind adding Changelog and Readme additions, then i will merge.

timhallinflux:
fixes #295

PR #358, opened 2017-08-01
Update BatchPoints.java

salmanjalali:
Added an option similar to Point.java to add multiple tags at once to the batch points.


majst01:
Hi,

A proper MR does at least have the following:
- based on a issues which describes the problem you are facing
- includes a bunch of unit- and integration-tests
- does at least compile

none of these minimal criteria's is met, so please think again.

PR #384, opened 2017-11-20
Migrate to JUnit5

majst01:
I would like to switch to the most recent version of junit, this is the first shot.

@fmachado would you please review ?

### DIFF
--- a/src/test/java/org/influxdb/InfluxDBTest.java
+++ b/src/test/java/org/influxdb/InfluxDBTest.java
@@ -306,9 +306,9 @@ public class InfluxDBTest {
         QueryResult result = this.influxDB.query(query);
 
         Assertions.assertEquals(3, result.getResults().get(0).getSeries().size());
-        Assertions.assertEquals(result.getResults().get(0).getSeries().get(0).getTags().get("atag"), "test1");
-        Assertions.assertEquals(result.getResults().get(0).getSeries().get(1).getTags().get("atag"), "test2");
-        Assertions.assertEquals(result.getResults().get(0).getSeries().get(2).getTags().get("atag"), "test3");
+        Assertions.assertEquals("test1", result.getResults().get(0).getSeries().get(0).getTags().get("atag"));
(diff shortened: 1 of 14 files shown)


majst01:
Code Coverage change far too much, which i don't understand. Maybe junit5 generates different coverage files ?

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -323,10 +326,10 @@ public void testWriteMultipleStringDataLinesThroughUDP() {
         Query query = new Query("SELECT * FROM " + measurement + " GROUP BY *", UDP_DATABASE);
         QueryResult result = this.influxDB.query(query);
 
-        Assert.assertEquals(3, result.getResults().get(0).getSeries().size());
-        Assert.assertEquals(result.getResults().get(0).getSeries().get(0).getTags().get("atag"), "test1");
-        Assert.assertEquals(result.getResults().get(0).getSeries().get(1).getTags().get("atag"), "test2");
-        Assert.assertEquals(result.getResults().get(0).getSeries().get(2).getTags().get("atag"), "test3");
+        Assertions.assertEquals(3, result.getResults().get(0).getSeries().size());
+        Assertions.assertEquals(result.getResults().get(0).getSeries().get(0).getTags().get("atag"), "test1");
```
fmachado: (minor adjustment) the "expected value" should be the first parameter

fmachado:
Codecov report seems to be correct now (it was updated 1h ago).

src/test/java/org/influxdb/InfluxDBTest.java
majst01: good  catch will fix.

majst01:
@fmachado should we first release 2.8 and postpone this into 2.9, what do you think ?

fmachado:
IMO we could add it right now as it will only affect the existing PR (well, most of them must be fixed or discarded anyway). The other positive thing is that next PRs will already be created using junit5.

majst01:
Agree, please merge it.

PR #400, opened 2017-12-16
fix #365 - fix problem of connecting to the influx api with URL which…

### LINKED ISSUE #365: Doesn't support urls not under '/'
Hi,
I'm using a JMX wrapper [jmxtrans](https://github.com/jmxtrans/jmxtrans) that uses this library to push data to influx.
Unfortunately my influxdb is not mapped under '/', so the URL is something like `https://xyz.com/influx/`. I tracked down the issue until this library, which appears to use all URLs directly under '/'.
Example in [InfluxDBService.java](https://github.com/influxdata/influxdb-java/blob/master/src/main/java/org/influxdb/impl/InfluxDBService.java) line 40:
```
  @GET("/query")
  public Call<QueryResult> query(@Query(U) String username, @Query(P) String password, @Query(DB) String db,
      @Query(EPOCH) String epoch, @Query(value = Q, encoded = true) String query);
```
This causes the library to fail because it defaults to '/' urls (which don't exist in my case):

```
java.lang.RuntimeException: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL /query was not found on this server.</p>
</body></html>
```
I couldn't find a way around this.
Thanks for any help!

vicctor:
… does not points to the url root (e.g. localhots:80/influx-api/)

This change was tested with following URLs:
http://localhost:8086
http://localhost:8086/

Proxied connection (influx behind Nginx)
http://loocalhost:8080/influx-api/


fmachado:
@vicctor would be great if you could provide a docker-compose.yml file containing both nginx properly configured and influxdb running so I could run the tests against it.

majst01:
Hi @vicctor 

Please do not format any of the files for a PR, try to reduce the changes needed, you have actually add more than 1000 lines and remove about 800, this makes it impossible to review this PR.

vicctor:
@majst01 Hi Steffan, thx for quick response. The truth is that I'm new in this project and I was in hurry to fix this issue, so all I can say sorry about that.
Indeed I found that I committed unnecessary white spaces in the pom file. I will reduce it and send yet another pull request later. Please, can you also give me some hints if I can reduce something more? It would be also great if you can provide me with coding-rules used in the project, so I can configure my IDE using it.

Also small explanation about what I did:
I added automatic tests with nginx proxy using the compile-and-test.sh (I found this approach is used at the moment on travis).
I can see however that this approach generates falsy codecov results. I'ts because tests are not started from surefire but from the .sh script. My first approach was to use the old good Parametriesed test and provide two different URI's to the InfluxDBTest class. But it turns out that with the JUnit5 I have to do that on every single method. Yesterday I did not found any more time to do that so perhaps it's good thing for next refac. Anyway, I'm also not sure if this is the best choice. Please let me know about your guess/plan.

MaxYuen:
thanks for ur contributions

majst01:
@vicctor if you still interested in this PR you need to rebase or  resolve the conflicts.

majst01:
This PR does basically the same as #416

pom.xml
```diff
@@ -89,6 +89,11 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
           <version>2.21.0</version>
+          <configuration>
+            <excludes>
+              <exclude>${someModule.test.excludes}</exclude>
```
majst01: What is this meant for

src/test/java/org/influxdb/InfluxDBFactoryTest.java
```diff
@@ -1,13 +1,12 @@
 package org.influxdb;
 
+import okhttp3.OkHttpClient;
 import org.influxdb.dto.Pong;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 
-import okhttp3.OkHttpClient;
```
majst01: Why reformatting this import

src/test/nginx/nginx.conf
```diff
@@ -0,0 +1,56 @@
...
+    #error_page  404              /404.html;
+
+    # redirect server error pages to the static page /50x.html
+    #
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+
+    # proxy the PHP scripts to Apache listening on 127.0.0.1:80
```
majst01: A lot of unrelated and commented code ?

lxhoan: @majst01, I am going to help Mr. @vicctor if he has no time for this PR, just trying to make this PR able to merge. I will address your comment soon

lxhoan:
@majst01, @fmachado . I improved the original works, please review this again, thank you

src/test/java/org/influxdb/InfluxDBProxyTest.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Test the InfluxDB API.
+ *
+ * @author hoan.le [at] bonitoo.io
+ *
+ */
+@RunWith(JUnitPlatform.class)
+public class InfluxDBProxyTest {
+  private InfluxDB influxDB;
+  private String db = "udp";
```
majst01: The database should be named via-proxy or so, but not udp, this is misleading

src/test/java/org/influxdb/TestUtils.java
```diff
@@ -8,38 +8,41 @@
...
-			if (getenv.containsKey("INFLUXDB_PORT_COLLECTD")) 
-				port = getenv.get("INFLUXDB_PORT_COLLECTD");
+			return getEnv("INFLUXDB_PORT_COLLECTD", "8096");
 		}			
-		
-		return port;
 	}
 
+	public static String getProxyApiUrl() {
+    return getEnv("PROXY_API_URL", "http://127.0.0.1:8086/");
```
majst01: can you please check formatting, here identation is broken

lxhoan: fixed it, we should have check style for test as well

src/test/java/org/influxdb/InfluxDBProxyTest.java
lxhoan: change naming
BTW, name of the database for UDP is configured at server side

PR #403, opened 2017-12-21
Add convenience method to Pong for checking ping status.

egoebelbecker:
Added method. Fixed checkstyle checks.

### DIFF
--- a/src/main/java/org/influxdb/dto/Pong.java
+++ b/src/main/java/org/influxdb/dto/Pong.java
@@ -26,9 +26,11 @@ public class Pong {
   }
 
   /**
-   * Good or bad connection status
+   * Good or bad connection status.
    */
-  public boolean isGood( ) { return !version.equalsIgnoreCase("unknown"); }
+  public boolean isGood() {
+    return !"unknown".equalsIgnoreCase(version);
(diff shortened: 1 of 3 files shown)


egoebelbecker:
Not sure if my changes caused these failures? It doesn't seem like they should have?

src/main/java/org/influxdb/dto/Pong.java
```diff
@@ -25,6 +25,13 @@ public void setVersion(final String version) {
     this.version = version;
   }
 
+  /**
+   * Good or bad connection status.
+   */
+  public boolean isGood() {
+    return !"unknown".equalsIgnoreCase(version);
```
fmachado: @egoebelbecker would be possible to transform _"unknown"_ into a class scoped `static final` constant?

fmachado:
CI failed for another reason that is not related with your changes.

src/main/java/org/influxdb/dto/Pong.java
egoebelbecker: Good point. I have no idea why I didn't do that in the first place. Cancelling this and submitting a new one.

PR #404, opened 2017-12-25
Guava20 upgrade

smunjal:


majst01:
Hi @smunjal 
This PR seems to be based on a very old revision of influxdb-java, it does not even apply. So please think again what you want to achieve.

Issue #399, opened 2017-12-16
Deprecate helper methods for later removal


fmachado:
`org.influxdb.InfluxDB` client interface provide some helper methods for database management that are incomplete and don't provide any concrete benefit to the developer: you can achieve the same result using `org.influxdb.InfluxDB.query(Query)` method.

I would like to avoid a library with tons of useless methods and the nightmare that will be our lives once we face the InfluxDB database API evolving and we having to track and be compatible with all versions (e.g. CREATE DATABASE for [v0.9](https://docs.influxdata.com/influxdb/v0.9/query_language/database_management/#create-a-database-with-create-database) is different than [v1.4](https://docs.influxdata.com/influxdb/v1.4/query_language/database_management/#create-database)). Also, it will avoid bugs like #398.

That said, here are the methods that I'm voting to deprecate for later removal:
* org.influxdb.InfluxDB.createDatabase(String)
* org.influxdb.InfluxDB.deleteDatabase(String)
* org.influxdb.InfluxDB.describeDatabases()
* org.influxdb.InfluxDB.databaseExists(String)
* org.influxdb.InfluxDB.createRetentionPolicy(String, String, String, String, int, boolean)
* org.influxdb.InfluxDB.createRetentionPolicy(String, String, String, int, boolean)
* org.influxdb.InfluxDB.createRetentionPolicy(String, String, String, String, int)
* org.influxdb.InfluxDB.dropRetentionPolicy(String, String)

If you have any good reason to keep them, please comment here.

ivanscattergood:
Hi, 
I think we should keep these functions, i.e. revert this change. 

I used to have these in a separate library myself and migrated over to the native tools when they were made available.

PR #407, opened 2018-01-03
Added 'deprecated' annotations to javadoc and methods

fmachado:
Implements #399 .
@majst01 In addition to
```Java
   /*
   * ... 
   * @deprecated (since 2.9) Use <tt>org.influxdb.InfluxDB.query(Query)</tt> to execute a parameterized
   *             <strong>CREATE DATABASE</strong> query.
   */
```
should I make a note like
```Java
   /*
   * ... 
   * @deprecated (since 2.9) Use <tt>org.influxdb.InfluxDB.query(Query)</tt> to execute a parameterized
   *             <strong>CREATE DATABASE</strong> query. Note: this method will be removed on the next major version.
   */
```

### DIFF
--- a/src/main/java/org/influxdb/InfluxDB.java
+++ b/src/main/java/org/influxdb/InfluxDB.java
@@ -342,7 +342,10 @@ public interface InfluxDB {
    *
    * @param name
    *            the name of the new database.
+   * @deprecated (since 2.9) Use <tt>org.influxdb.InfluxDB.query(Query)</tt> to execute a parameterized
+   *             <strong>CREATE DATABASE</strong> query.
    */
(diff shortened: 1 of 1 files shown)


majst01:
Hi @fmachado 
Yes i think we should clearly state when (in which version) these methods are removed. I would prefer a `git grep` friendly reminder notice to get all methods to remove in one go.

possible format:

```java

/*
   * ... 
   * @deprecated (since 2.9, removed in 3.0) Use <tt>org.influxdb.InfluxDB.query(Query)</tt> to execute a parameterized
   *             <strong>CREATE DATABASE</strong> query.
   */

```

suggestions welcom

PR #409, opened 2018-01-09
enableBatch accepts parameters as BatchOptions class

dubsky:
This pull request implements a change in how enableBatch is called - instead of having many parameters to the function, it just receives an object. 

It introduces new configuration property 'jitter interval' to avoid multiple clients hit the server periodically at the same time. (With the same buffer flush interval and started at the same time)


src/main/java/org/influxdb/BatchOptions.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.concurrent.ThreadFactory;
+import java.util.function.BiConsumer;
+
+/**
+ * BatchOptions are used to configure batching of individual data point writes
+ * into InfluxDB. See {@link InfluxDB#enableBatch(BatchOptions)}
+ */
+public interface BatchOptions {
+
+  BatchOptions DEFAULTS = BatchOptionsImpl.DEFAULTS;
```
majst01: This is kinda weird to have a backreference to the impl in the interface

src/main/java/org/influxdb/impl/BatchProcessor.java
```diff
@@ -196,15 +216,16 @@ public static Builder builder(final InfluxDB influxDB) {
     return new Builder(influxDB);
   }
 
-  private BatchProcessor(final InfluxDBImpl influxDB, final ThreadFactory threadFactory, final int actions,
```
majst01: Why is the weaker signature needed ?

src/main/java/org/influxdb/BatchOptions.java
majst01: Why a interface for the options at all, decoupling is not achieved, why the extra effort ?

majst01:
@dubsky Please add proper documention as well, we will otherwise face a lot of issues from people using this library and wonder how it is ment to work. Your Javadoc in the interface is a good starting point, a sample implementation would be helpful.

src/main/java/org/influxdb/impl/BatchProcessor.java
dubsky: This is because I have been using more strict rules for check style at first by mistake. Sun style wants it like this. I don't think there is a real difference the only thing I can imagine it may affect javadoc at certain settings.

src/main/java/org/influxdb/BatchOptions.java
dubsky: I wanted to have javadoc documentation in a separate file and this thing is best I could do. Perhaps I may merge the code from the interface and the class into one BatchOptions class. Defaults would be reachable be BatchOptions.DEFAULTS. You think that this way it would be better ? Or do you have some other idea ?

dubsky:
You mean updating the README.md ? I didn't rush into it before we agree this approach is the way to go. Also, I would consider deprecating most if not all enableBatch methods that consume the isolated configuration properties. That would also mean to rework the documentation so that it uses the new configuration method.

src/main/java/org/influxdb/BatchOptions.java
majst01: Yes i would prefer this. I think we have no benefit have splitting Batchoptions into Interface and Implementation.

src/main/java/org/influxdb/impl/BatchProcessor.java
majst01: OK

majst01:
Regarding documentation i would prefer to have the documentation either in the Readme.md or in a seperate document which is referenced from the Readme.md. But either way it should come from a this PR to have a clear reference when this new functionality was added.

dubsky:
@majst01 The failed test here is PerformanceTests.testWriteCompareUDPPerformanceForBatchWithSinglePoints. It is not a problem of this PR but rather the test itself. (Actually, it asserts that String.join of 1000 strings takes 1ms at least) It doesn't make much sense in my point of view and I would recommend to remove it.

dubsky:
@majst01 @fmachado I have updated the documentation - I have also change the concept a bit - batching enabled is the first example the user is going to see because I think it should be most commonly used. Perhaps it will reduce the number of people asking how to use the client asynchronously.

About the code coverage  - there is a suite of tests for BatchOptions ready inside #410. If you won't mind we will keep it there.

majst01:
Looks good to me, @dubsky would you please add a proper CHANGELOG.md entry afterwards, thanks.

PR #410, opened 2018-01-09
Improve error handling and implement data point write retry if makes sense

dubsky:
This pull request implements a feature where clients try to retry a write of DataPoints into Influx when the previous write failed and there is a chance that retry will be successful. The behavior and default settings are consistent with Telegraf default settings and behaviour.

There is a new BatchWriter interface with two implementations: OneShotBatchWriter which is the original behavior and RetryCapableBatchWriter which is the new one. To use the feature the InfluxDB.enableBatch(BatchOptions) constructor must be used.


majst01:
This PR overlaps with #409, so review can only work if first PR is approved and merged.

You added an extra dependency to org.json, but im almost sure that with retrofit we already have a dependency to a json decoder/encoder already. This needs to be checked to keep the library small and fast.

majst01:
The bundled json library in use is: https://github.com/square/moshi

dubsky:
Ok, I will use this one.

src/main/java/org/influxdb/InfluxDBException.java
```diff
@@ -18,4 +18,111 @@ public InfluxDBException(final String message, final Throwable cause) {
...
     super(cause);
   }
+
+  /**
+   * @return true if the operation may succeed if repeated, false otherwise.
+   */
+  public boolean isRetryWorth() {
+    return true;
+  }
+
```
fmachado: Could you please add here a comment pointing to where these messages can be found (e.g. InfluxDB source files, documentation, etc)? This would help the next developer who may need to work on this code.

majst01:
@dubsky @lxhoan please do not abuse travis to run tests on every single commit, do your work locally and test there (i provided the needed infrastructure) before committing in so many small modifications.

src/test/java/org/influxdb/impl/RetryCapableBatchWriterTest.java
```diff
@@ -0,0 +1,89 @@
...
+    Exception nonRecoverable = InfluxDBException.buildExceptionForErrorState("database not found: cvfdgf");
+    Exception recoverable = InfluxDBException.buildExceptionForErrorState("cache-max-memory-size exceeded 104/1400");
+    Mockito.doThrow(nonRecoverable).when(mockInfluxDB).write(bp0);
+    Mockito.doThrow(recoverable).when(mockInfluxDB).write(bp1);
+    Mockito.doThrow(recoverable).when(mockInfluxDB).write(bp2);
+    Mockito.doThrow(recoverable).when(mockInfluxDB).write(bp3);
+    // first one will fail with non-recoverable error
+    rw.write(Collections.singletonList(bp0));
+    // second one will fail with recoverable error
+    rw.write(Collections.singletonList(bp1));
```
fmachado: Let's suppose that `bp1` is failing with recoverable error and `bp2` is writing successfully (i.e. WriteResultOutcome.WRITTEN).
If both `bp1` and `bp2` are writing to the same point in time (but `bp2` contains the most recent values), I think the current retry logic will write `bp1` and overwrite the existing values with old ones. Am I wrong?

fmachado: Just to be clear:
`bp1`
```
time                 sometag somevalue
----                 ------- ----
2017-06-20T15:32:46Z foo     ON
2017-06-20T15:32:47Z foo     OFF
2017-06-20T15:32:48Z foo     ON
```

`bp2`
```
time                 sometag somevalue
----                 ------- ----
2017-06-20T15:32:48Z foo     OFF
2017-06-20T15:32:49Z bar     OFF
2017-06-20T15:32:50Z bar     OFF
```
`2017-06-20T15:32:48Z foo OFF` is the most recent point but as we are retrying `bp1`, it will be overwritten with `2017-06-20T15:32:48Z foo ON`

dubsky: @fmachado No, under usual circumstances the implementation won't try to write bp2 after bp1 write failed on a recoverable error, it will just get queued in the RetryCapableBatchWriter. The case you describe might happen when the bufferLimit gets exceeded, in that case oldest batch points that failed on a recoverable error get forgotten.

This test is a junit test targetted on the RetryCapableBatchWriter internals - there are no writes into InfluxDB happening right now, that is why the conflicts in timestamps between bp1 and bp2 don't matter. Hoan is building BatchOptionsTest for which the implementation will be a black box.

lxhoan: RetryCapableBatchWriter follows the approach of telegraf, which is aware of and guarantees the chronological order of the written batchpoints, even when bufferLimit is exceeded.
For example, with Fernando's case, RetryCapableBatchWriter will make sure bp1 will only be rewritten before the writing of bp2
I think for now we only need a test in RetryCapableBatchWriterTest for this logic. I added RetryCapableBatchWriterTest.testRetryingKeepChronologicalOrder()

src/main/java/org/influxdb/InfluxDBException.java
dubsky: It is now documented in the InfluxDBException class source code.

majst01: But every developer would stumble over this code because he would ask himself what this method is all about. Is there any future modification expected that this method would return something different than true in the near future ? If no, please remove.

dubsky: There is a couple of subclasses of InfluxDBException for specific errors that return a different value in isRetryWorth() already. For example AuthorizationFailedException, DatabaseNotFoundException. I also added a javadoc comment for explanation.

lxhoan:
@fmachado @majst01 @dubsky 
To conduct a black-box testing for this new implementation (the retrying capability). I tried this InfluxDB configuration
image

The used EC2 instance it t2.micro, the purpose is to scale down the DB server so we can overload it more easily

Then I run the PerformanceTests.testMaxWritePointsPerformance(), what I observed:

+ In master branch: the test failed with SocketTimeoutException (DB server threw a lot of errors)
+ In error-handling branch (of course a bit modification to use BatchOptions) : the test is not failed (it just stands still and waiting for the retrying). The client behaves better (similarly to Telegraf), no exception thrown, batch retrying works (though DB server still spits out similar errors. But I believe PerformanceTests.testMaxWritePointsPerformance() exceeds to much the capability of this scaled-down DB configuration) 

In case we use this new java client for an (intermittently) busy DB server, it will help a better result

majst01:
@lxhoan great to hear, would be perfect to express this in a travis ci integration test. Should be documented at travis howto do so.

lxhoan:
There some documents used internally (the InfluxDB for testing is enterprise version), I will check and reproduce them for travis how-tos.

fmachado:
> In error-handling branch (of course a bit modification to use BatchOptions) : the test is not failed (it just stands still and waiting for the retrying). The client behaves better (similarly to Telegraf), no exception thrown, batch retrying works (though DB server still spits out similar errors. But I believe PerformanceTests.testMaxWritePointsPerformance() exceeds to much the capability of this scaled-down DB configuration)

So, did you confirm that no points were lost?

lxhoan:
> the test is not failed (it just stands still and waiting for the retrying)
That is what I observed
Batch writing failed by recoverable error would be put to retry buffer. In this black box scenario the DB server could not be back to normal so retrying kept unsuccessful
The purpose of this test is to verify if the influxdb-java can behave as same as Telegraf (that's the expectation)
As I said 
> In case we use this new java client for an (intermittently) busy DB server, it will help a better result
That means if you would like to verify the efficiency of this new feature, you will need to make a temporarily unstable/overloading state for the DB server, it's feasible but not an immediately easy task, especially for CI integration test

fmachado:
Ok, If I got it right you executed a test and confirmed that the "retry mechanism" is working because your test kept running (maybe throwing exceptions) even when the InfluxDB was (almost completely) non-responsive due to extreme load. Is this correct?

lxhoan:
yes, it's correct. I think you have what I meant

fmachado:
Would be possible to change a little bit the way we test this new feature?
1) Identify when the t2.micro instances begin to have issues handling the requests;
2) Identify when the t2.micro instances are completely overloaded and InfluxBD probably would take **a lot of time** to recover;

**Why?**
IMO we should perform this test on a recoverable scenario (1) and avoid (2). If we can do this then my initial question can be answered: *"can you confirm that no points were lost?"*

If you try to write 1M points respecting (1) then you should be able to, at least, manually execute a SELECT COUNT on InfluxDB.

Is this feasible or do you have plans to this kind of test later?

lxhoan:
If I understood well then by (1) you mean my comment for "a temporarily unstable/overloading state" ? If so then the answer was mentioned already: yes, it's possible but not too easily.

fmachado:
If you have access to Amazon Cloud Watch, you may check the "CPU credits" that your t2.micro instances have as they are "[burstable](https://aws.amazon.com/ec2/instance-types/#burst)" in conjunction with the load. Check this [doc](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/t2-credits-baseline-concepts.html) for more information.

If you don't have access to it or don't want to use it as a reference, no problem. What do you think about trial and error? :) Try to write 100K points... Was this enough to bring load to the t2.micro instances? If not, keep doubling until you get the right amount of data.

Would this be easier? What do you think?

lxhoan:
I just read the document you provided, but it's not clear how to dynamically scale the t2.micro CPU performance the way we want. Do you have any hint on this ?

majst01:
It could be much easier if we use a influxdb container during ci, which is what we already do, and limit for this type of integration test a capped amount of cpu resources for the influxdb container. CPU Limits can be specified at docker run as described here: https://docs.docker.com/config/containers/resource_constraints/#cpu 

So we should add a second travis job which is solely for integration test and limit cpu for the influxdb container accordingly. So no additional resources are required at AWS et.al.

lxhoan:
I am actually using docker for my test configuration (picture above). What I am thinking is a **dynamically** and **programmably scaling** mechanism for the CPU performance of the host machines (t2.instance or docker container or whatever) to make sure we can automate the whole integration test script. I am not sure if docker provides such a mechanism, I will need to check and decide an acceptable solution, any idea is welcome. This may take time so I suggest to consider it as a separate plan from this pull request, what do you think ?

BTW from unit testing approach, it's easy to stub the InfluxDBImpl and script it to simulate the overloading state of DB server such as always throw connection timeout for a specific amount of time. So we can make sure the retrying really works and no point lost.

majst01:
What is important for me, all effort must be public visible, that said a integration test must at the end be implemented in the current travis configuration that whatever PR we get in the future will always be checked against we have, losing visibility is not acceptable an will lead to unmaintainable code.

@lxhoan what you actually want to achieve should be possible with docker as well, no extra vm's needed. Please have a look into https://github.com/influxdata/influxdb-java/blob/master/compile-and-test.sh how actual unit testing works.

dubsky:
I have a feeling that this is going a little bit sideways. First, I think that the test Hoan made was manual only just to double check our effort. There are automatic tests for the retry behavior already and I think that the PR is ready for merge.

I think that it is important to separate testing of influxdb-java client and influxdb server itself and avoid the later one within influxdb-java tests because it would be difficult either to implement it and later on maintain it. I think that for influxdb-java tests for various difficult to arrange scenarios it is perfectly ok to virtualize (simulate) the influxdb behavior.

As we talk about the docker cpu resource constraints I am not sure it is a good idea  - the resource constraints parameters are given by fractions of the CPU the docker is running on. You would be given completely different performace when running the same docker image with exactly the same resource constraints on two different computers with different CPU performance. And that can be quite a difference - my workstation is in the influxdb-java test 20x-50x more powerful than Travis CI. This gap would also make the development of a test based on CPU throttling very difficult. (Also, what happens if Travis starts to use new HW...)

fmachado:
@dubsky I understand that Hoan was trying to double check the implementation and I really appreciated his effort. Maybe for the next time we can find a better way to do these kind of tests. IMHO, an application throwing exceptions with an non-responsive backend is not an acceptable test.
I already approved the merge. Let's wait for @majst01.

ivankudibal:
@fmachado - for future releases we should consider how to test enterprise influxdb, we can use corporate AWS account to get physical hw.. and test executor will need to report back (as new comment in PR, implemented similarly to travis-ci or codecov services)...  for those purposes I created the bonitoo-builder account that can trigger an event and contribute with test report...

majst01:
@dubsky i have concerns that this quite important feature is left without proper tests once merged. This is why i would love to see at least that all the points are written if the database is on a overload situation and no points are lost or doubled by the retry algorithm. I am sure this can be done with a cpu restricted influxdb running in a container serving requests for a integration test. I am happy to support @lxhoan to implement this. 

@ivankudibal testing against the enterprise edition of influxdb implies that this version does behave different from the small api we are facing, i hope that is not the case. If so, can you please tell what different behavior do you expect.

lxhoan:
>BTW from unit testing approach, it's easy to stub the InfluxDBImpl and script it to simulate the overloading state of DB server such as always throw connection timeout for a specific amount of time. So we can make sure the retrying really works and no point lost.
I added org.influxdb.PerformanceTests.testRetryWritePointsInBatch(), it writes 400k point to the DB. The test stubs influxDB service to throw a retrying capable exception for the first 80s, then back to normal, the purpose to simulate an unstable DB state for 80s. All 400k point writing is failed for few times but after that 80 s, they are all written successfully to DB (by the retrying capability feature)

majst01:
@lxhoan great, even simpler i like it.

PR #414, opened 2018-01-24
fix Vlada's comments

lxhoan:


dubsky:
testConsitency()  - there is still no assertion testing that the consistency value set is propagated to InfluxDB. You need to dig deeper into the implementation (use the debug interceptor for that ?) It is obvious it is in BatchOptions (there is a test on that already).

dubsky:
testBufferLimitGreaterThanActions()
testBufferLimitLessThanActions()

You need to assert that retry is happening or it is not happening for the test to be meaningful.

PR #416, opened 2018-01-31
Support get/post requests behind http reverse proxy like nginx and ba…

essagl:
…sic authorization.


src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -53,7 +53,7 @@
  */
 public class InfluxDBImpl implements InfluxDB {
 
-  static final okhttp3.MediaType MEDIA_TYPE_STRING = MediaType.parse("text/plain");
+  static final MediaType MEDIA_TYPE_STRING = MediaType.parse("text/plain");
```
fmachado: IMO, the previous code was a bit more clear by stating that `MediaType` comes from `okhttp3` and not from another package (e.g. `javax.ws.rs.core.MediaType`).
I don't like also the name as there is no media type string (accordingly with the RFC) but this is another issue.

fmachado:
@essagl, I would like to proceed reviewing this PR but there are some points that I think should be important to mention:
1) There is no "Issue ticket" created explaining the problem you are trying to solve. If you had done this before creating this PR, someone would have pointed to you that you can pass a `OkHttpClient.Builder` parameter with your own `okhttp3.Authenticator` implementation;
2) You are assuming the same `username` and `password` to access a database will be used on the HAProxy side. This may be OK for a single user<->database instance but what would happen if I have a InfluxDB with multiple databases and multiple users?

IMO, unless you have a clever way to fix (2) I think this PR is invalid and won't be accepted. :(

essagl:
Thank you for your reply,

to 1) i faced the same problem as described in
https://github.com/influxdata/influxdb-java/issues/268 which would be fixed
now.
to 2) as mentioned in the documentation at
https://docs.influxdata.com/influxdb/v1.4/query_language/authentication_and_authorization/
the prefered method is Authenticate with Basic Authentication as described
in RFC 2617, Section 2. <http://tools.ietf.org/html/rfc2617>
So i thought, that there would never be a problem if the proxy will just
let pass the authorization header.
It will be up to the user to decide if the proxy or the database should
authenticate.
So the "clever way to fix (2)" could be simply to not configure basic
auhorization with your proxy.

#essagl



<http://tools.ietf.org/html/rfc2617>

2018-01-31 13:13 GMT+01:00 Fernando Machado <notifications@github.com>:

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -429,11 +436,10 @@ public void onResponse(final Call<ResponseBody> call, final Response<ResponseBod
                 try {
                     if (response.isSuccessful()) {
                         BufferedSource source = response.body().source();
-                        while (true) {
-                            QueryResult result = InfluxDBImpl.this.adapter.fromJson(source);
-                            if (result != null) {
-                                consumer.accept(result);
-                            }
+                        QueryResult result = InfluxDBImpl.this.adapter.fromJson(source);
```
majst01: Why is this implementation changed, seem unrelated to what you want to achieve.

src/main/java/org/influxdb/impl/InfluxDBService.java
```diff
@@ -1,9 +1,8 @@
 package org.influxdb.impl;
 
-import org.influxdb.dto.QueryResult;
-
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
+import org.influxdb.dto.QueryResult;
```
majst01: In general do not reformat the code in parts which are not modified.

src/main/java/org/influxdb/impl/InfluxDBService.java
```diff
@@ -37,35 +36,36 @@
    * @param consistency optional The write consistency level required for the write to succeed.
    *                             Can be one of one, any, all, quorum. Defaults to all.
    */
-  @POST("/write")
+  @POST("write")
   public Call<ResponseBody> writePoints(@Query(U) String username,
-      @Query(P) String password, @Query(DB) String database,
-      @Query(RP) String retentionPolicy, @Query(PRECISION) String precision,
-      @Query(CONSISTENCY) String consistency, @Body RequestBody batchPoints);
+                                        @Query(P) String password, @Query(DB) String database,
```
majst01: same here, and all other parts down here.

src/main/java/org/influxdb/impl/RequestAuthenticator.java
```diff
@@ -0,0 +1,39 @@
...
+
+import okhttp3.Authenticator;
+import okhttp3.Credentials;
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.Route;
+
+import java.io.IOException;
+
+public class RequestAuthenticator implements Authenticator {
```
majst01: No a single test covering this new code, without this we cant accept this PR.

majst01:
I totally agree @fmachado, PR in this form is not going to be accepted.

src/main/java/org/influxdb/impl/InfluxDBService.java
essagl: Noticed, that codestyle check stops building the project when there are lines longer tan 120 characters

src/main/java/org/influxdb/impl/RequestAuthenticator.java
essagl: Would a JUnit test be sufficient? Like this one:
https://github.com/essagl/influxdb-java/blob/master/src/test/java/org/influxdb/impl/RequestAuthenticatorTest.java

src/main/java/org/influxdb/impl/InfluxDBImpl.java
essagl: IntelliJ IDE showed an error because of an endless loop

mwmahlberg:
@umetzger Furthermore, I'd like to object regarding the authentication issue. In high security environments, it may well be that the client runs in a DMZ and the reverse proxy serves as an application level gateway, requiring authentication as an additional level of security. So both authentication against the proxy and InfluxDB would be an actual requirement. Looking at the environments  InfluxDB is likely to be used in, I'd strongly recommend and kindly ask to reconsider this point.

mwmahlberg:
@umetzger Sorry, last comment was for OP.

majst01:
@essagl can you fix the merge conflicts that we can go ahead with the merge process.

compile-and-test.sh
```diff
@@ -3,9 +3,9 @@
 # script to start influxdb and compile influxdb-java with all tests.
 #
 set -e
-
+test -t 1 && USE_TTY="-it"
 DEFAULT_INFLUXDB_VERSION="1.5"
-DEFAULT_MAVEN_JAVA_VERSION="3-jdk-10-slim"
+DEFAULT_MAVEN_JAVA_VERSION="3-jdk-9-slim"
```
majst01: this should stay at jdk10

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -492,8 +499,7 @@ public void onResponse(final Call<ResponseBody> call, final Response<ResponseBod
                             if (result != null) {
                                 consumer.accept(result);
                             }
-                        }
-                    }
+                        }                    }
```
majst01: Formatting

compile-and-test.sh
essagl: Fixed

src/main/java/org/influxdb/impl/InfluxDBImpl.java
essagl: ok

lxhoan:
Hi @majst01, this PR looks OK now, will you merge it ?

majst01:
I am still not confident, the Issue 2) @fmachado raised regarding authentication is for me valid and still unanswered. @essagl can you elaborate on that topic ?

essagl:
@majst01 the Issue 2) @fmachado raised regarding authentication is for me valid and still unanswered.

Securing access to influxdb throught basic authentication done by the proxy is optional and just another easy way to restict access to an insecured influxdb. This can be a convinient approach for those who already have a http proxy in place. (This is also the usecase which is covered by the tests.) 
As @fmachado sayed this will also work if the proxy user/password is the same as the influxdb user/password.
So in more complex authentication scenarios one solution would be, not to configure your proxy with basic authentication for your influxdb location. The influxdb itself is than responsible for this concern, as it would be the case if no proxy is installed at all.

lxhoan:
According to the document (@essagl  shown), InfluxDB server supports:
   + basic auth (preferred)
   + parameter auth (takes precedence if both methods provided to http request)

currently influxdb-java client uses parameter auth

@essagl, I think you can split your Basic auth impl (RequestAuthenticator) to a separate PR (I suggest also open a new issue ticket for tracking), it's quite fair to open such a PR because the official document states that Basic Auth is the preferred method. I hope this approach will satisfy both your expectation and @fmachado 's concern

lxhoan:
I open this kind/enhancement ticket issue #449

lxhoan:
add label 'question'
remove from milestone 2.11 since no response from PR author @essagl so far

essagl:
@*lxhoan <https://github.com/lxhoan>* sorry, but at the next few weeks
there is no time for me to deal with the influxdb client. The projects i
needed the "basic auth behind nginx" feature for are satisfied with the
solution. I just wanted to share, so feel free to do whatever you want.


2018-06-01 11:22 GMT+02:00 Hoan Xuan Le <notifications@github.com>:

lxhoan:
closed this due to merging of PR #400 and PR #449

Issue #339, opened 2017-06-16
fromInfluxDBTimeFormat does not parse milliseconds correct


Prebzs:
If a timestamp has a 10-millisecond time (last digit 0), the returned milliseconds time is not correct.

.time(1497600000030L, TimeUnit.MILLISECONDS)

after reading from database: 
fromInfluxDBTimeFormat(raw.get(0).toString()) returns "1497600000003"

the database contains "2017-06-16T08:00:00.03Z"

fmachado:
TLDR: When you execute a query using the InfluxDB Java client, always specify the TimeUnit parameter.

This seems to be related with how InfluxDB (in fact, Go lang) uses the RFC3339 for date formatting and how _java.text.SimpleDateFormat_ (used by this library) differs from the first.

\<short-version\>
* InfluxDB format "1497600000030L" as "2017-06-16T08:00:00.03Z" (.030 ms was trimmed to .03 ms)
* _SimpleDateFormat_ using the pattern "yyyy-MM-dd'T'HH:mm:ss.SS'Z'" format "1497600000030L" as "2017-06-16T10:00:00.30Z" (.030 ms was trimmed to .30 ms)
* If you try to parse the "2017-06-16T08:00:00.03Z" using _SimpleDateFormat_, you will have **the wrong value with .03ms**.

\</short-version\>



You can test it by yourself:
```
root@3e65d276839c:/# influx                   
Connected to http://localhost:8086 version 1.3.0
InfluxDB shell version: 1.3.0
>
> create database mydatabase
>
> use mydatabase
Using database mydatabase
>
> insert bar value=1 1497600000030000000
>
> select * from bar
name: bar
time                value
----                -----
1497600000030000000 1
```
The time _1497600000030000000_ is in nanoseconds and it is the same one provided in the previous comment.

Now let's see how InfluxDB is formatting a date using [RFC3339](https://www.ietf.org/rfc/rfc3339.txt). Accordingly with the InfluxDB documentation:

> The -precision argument specifies the format/precision of any returned timestamps. In the example above, rfc3339 tells InfluxDB to return timestamps in RFC3339 format (YYYY-MM-DDTHH:MM:SS.nnnnnnnnnZ).
> Source: https://docs.influxdata.com/influxdb/v1.2/introduction/getting_started/#creating-a-database

```
root@3e65d276839c:/# influx -precision rfc3339
Connected to http://localhost:8086 version 1.3.0
InfluxDB shell version: 1.3.0
>
[...]

PR #418, opened 2018-02-11
implement query() call to make asynchronous InfluxDb query with TimeUnit

vkurland:
Hi

This change introduces a non-blocking query() call that also allows for the TimeUnit parameter to get timestamps in microseconds or nanoseconds rather than in RFC3339. Currently (in the upstream master) we either can do a non-blocking call (with onSuccess and onFailure callbacks) but get timestamps in RFC3339, or specify timestamp unit but then the call is blocking. This change provides a way to have both.

thank you
Vadim Kurland

### DIFF
--- a/src/main/java/org/influxdb/InfluxDB.java
+++ b/src/main/java/org/influxdb/InfluxDB.java
@@ -357,6 +357,22 @@ public interface InfluxDB {
   public void query(final Query query, final Consumer<QueryResult> onSuccess, final Consumer<Throwable> onFailure);
 
   /**
+   * Execute a query against a database with ability to specify the time unit.
+   *
+   * One of the consumers will be executed.
+   *
+   * @param query
+   *            the query to execute.
+   * @param timeUnit the time unit of the results.
+   * @param onSuccess
+   *            the consumer to invoke when result is received
+   * @param onFailure
+   *            the consumer to invoke when error is thrown
+   */
+  public void query(final Query query, TimeUnit timeUnit,
+                    final Consumer<QueryResult> onSuccess, final Consumer<Throwable> onFailure);
+
+  /**
    * Execute a streaming query against a database.
    *
    * @param query
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -438,6 +438,27 @@ public class InfluxDBImpl implements InfluxDB {
    * {@inheritDoc}
    */
   @Override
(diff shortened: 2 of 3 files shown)


majst01:
Can you please explain a bit more why this is needed. At least a unit test showing it works is required.

vkurland:
sure, happy to.

about the unit test - do I create a new pull request with the unit test included or can I add it to this one somehow? sorry for asking a stupid question, I do not do pull requests very often.

about the justification:

it comes down to the ability to satisfy two requirements: 1) be able to pass our own preconfigured http client for SSL auth and rate limiting and 2) to get timestamps in microseconds instead of as strings in RFC3339 format

your library uses Retrofit2 and so do we. Our application uses Inflixdb as a generic time series database, along with 3 more different TSDB implementations that we support (e.g. Graphite). It is possible for a user to run several instances of our application on-premise and all these instances can share the same Influxdb server. We need to be able to rate-limit queries these instances can send to Influxdb at the same time. Without this rate limiting, it is quite possible for one instance of our app to fire few thousand queries at the same time. Retrofit2 (or rather okhttp3 they use) provides a way to control the parallelism of the queries they make through the configuration of the Dispatcher. You have a variant of the query() call that accepts Builder which is a way to pass http client that we create and configure to the inflixdb library. We configure the Dispatcher used by our client to limit the number of simultaneous connections to the same host (just what we need).

One has to call retrofit2 enqueue() to utilize the rate limiting. Their call execute() is blocking and executes in the caller's thread so it does not use the Dispatcher. You can configure Dispatcher with a reasonable maxRequestsPerHost and still fire thousands of queries in parallel if you call execute() using your own thread pool.

Another important reason to use the variant of the query() call where we pass our own HTTP client is to be able to preconfigure HTTP client with SSL truststore and keystore for SSL auth when Influxdb runs with SSL

About the timestamps: we use Influxdb as a time series storage and the data we get from the query() call can be used to do various things before it is presented to the user. For example, we need timestamps in microseconds when we perform calculations with the data, or when we build graphs. Issue #339 explains difficulties with converting the time from RFC3339 to timestamps, besides, InfluxDB API has a way to request the time in microseconds anyway (or milliseconds or any other units). You have a variant of the query() call that accepts the parameter for that, but this call is blocking because it uses retrofit2 execute() instead of enqueue().

These are the reasons why query() call that can accept TimeUnit parameter and uses enqueue() to run actual InfluxDb API query is useful.

majst01:
Hi @vkurland if still interested please rebase and resolve the merge conflicts.

PR #421, opened 2018-02-27
Simplify the release

tomklapka:
Created a release profile which is automatically activated when running the maven-release-plugin including the InfluxDB started before tests execution (run via the docker-compose-maven-plugin) and stopped after.


majst01:
Guys please do not merge stuff which changes a whole file. Always check before you commit and merge at the end that the formatting is untouched. I dont like this this way because its impossible to see what actually changed. 

@ivankudibal please revert and fix formatting first.

majst01:
@ivankudibal @tomklapka still not reverted, please i really do not want to have such commits which change ~250 lines with no obvious view what changed.

tomklapka:
Fixed with commit @0e78b6c05cbdaeb8d1fd131761e16c53579d5c92

Issue #371, opened 2017-08-30
Support dynamic measurement name in InfluxDBResultMapper


dparmar:
In my case the measurement name is not known at compile time. So I was wondering if toPOJO() can take it as a parameter.

As a quick fix, currently I extend InfluxDBResultMapper and override getMeasurementName, but that doesn't look like elegant solution.

Thanks for this very useful feature.

fmachado:
@dparmar we could add a new parameter to `toPOJO` method where you could pass a measurement name. But before we do it, I'm curious: how do you name your measurements? Are you naming them using a pattern like `<measurement>_yyyymmdd` (e.g. `cpu_20170901`)?

dparmar:
They are user defined names like test_1, test_2

martinschuberts:
Why don't you use then a database or tag for it?
We for example have a database for each user - so we can clearly separate the data and access to it.
But the measurements are the same - just the tag values are different. 
I am curious about your use-case and approach.

ikercrg:
We have also the same problem. The solution suggested about adding a new parameter to `toPojo` would be great. 

We do not follow any specific pattern to name the measurements; sometimes it is name_subname or name.subname, etc..

Do you need more info, because this feature would be very interesting for us.

Thanks in advance

PR #423, opened 2018-03-01
#371 Support dynamic measurement name in InfluxDBResultMapper

ikercrg:
Checkstyle not added. Please review it. Added just one single method that supports a dynamic measurement name.

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
@@ -110,6 +110,8 @@ public class InfluxDBResultMapper {
    */
   public <T> List<T> toPOJO(final QueryResult queryResult, final Class<T> clazz, final String measurementName)
       throws InfluxDBMapperException {
+
+    Objects.requireNonNull(measurementName, "measurementName");
     Objects.requireNonNull(queryResult, "queryResult");
     Objects.requireNonNull(clazz, "clazz");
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
```diff
@@ -75,31 +75,63 @@
...
     Objects.requireNonNull(clazz, "clazz");
 
-    throwExceptionIfMissingAnnotation(clazz);
     throwExceptionIfResultWithError(queryResult);
     cacheMeasurementClass(clazz);
 
     List<T> result = new LinkedList<T>();
-    String measurementName = getMeasurementName(clazz);
+
+    if (measurementName == null || measurementName.isEmpty())
```
fmachado: Please move these verification to the top of the method. IMO `Objects.requireNonNull(measurementName, "measurementName");` would be enough and the code should be resilient against an empty string.

src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
```diff
@@ -75,31 +75,63 @@
...
+    {
```
fmachado: This block `{ }` is not compliant with the existing checkstyle.xml that you can find at the root of the project. If you have considered my previous argument as valid (Object.requireNonNull as enough), just remove this block.

src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
```diff
@@ -75,31 +75,63 @@
...
     queryResult.getResults().stream()
-      .filter(internalResult -> Objects.nonNull(internalResult) && Objects.nonNull(internalResult.getSeries()))
-      .forEach(internalResult -> {
-        internalResult.getSeries().stream()
-          .filter(series -> series.getName().equals(measurementName))
-          .forEachOrdered(series -> {
-            parseSeriesAs(series, clazz, result);
-          });
-      });
+        .filter(internalResult -> Objects.nonNull(internalResult) && Objects.nonNull(internalResult.getSeries()))
```
fmachado: Please, do not change the indentation of lines that are not direct related with your changes. Unfortunately you will have to revert these changes.

fmachado:
@ikercrg I would like to merge this PR but I cannot proceed due to the missing Unit Tests for the new method. Can you do it?

PR #424, opened 2018-03-09
Performance: Escape fields and keys more efficiently

mschaefers:
Under heavy load a profiler shows that calling `String.replace()` in `FIELD_ESCAPER` or `KEY_ESCAPER` is significantly slower than escaping manually by writing directly into the current `StringBuffer`.

This pull request does not change behavior but only execution speed

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -10,7 +10,6 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.TreeMap;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
 
 import org.influxdb.impl.Preconditions;
 
@@ -26,11 +25,6 @@ public class Point {
   private Long time;
   private TimeUnit precision = TimeUnit.NANOSECONDS;
   private Map<String, Object> fields;
-
-  private static final Function<String, String> FIELD_ESCAPER = s ->
-      s.replace("\\", "\\\\").replace("\"", "\\\"");
-  private static final Function<String, String> KEY_ESCAPER = s ->
-      s.replace(" ", "\\ ").replace(",", "\\,").replace("=", "\\=");
   private static final int MAX_FRACTION_DIGITS = 340;
   private static final ThreadLocal<NumberFormat> NUMBER_FORMATTER =
           ThreadLocal.withInitial(() -> {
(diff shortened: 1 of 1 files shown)


majst01:
Any numbers about performance improvements ?

mschaefers:
I used jvmtop (https://github.com/patric-r/jvmtop) as a simple profiler. Profiling our application resulted in 14-15% of CPU time to be used in closure `KEY_ESCAPER`. With this patch the CPU percentage of `escapeKey` was reported with 1-2%.

mschaefers:
Another very simple test can be done by executing the following testcase in `PointTest.java`

It creates 2M different lineProtocol Strings in a loop of 1M, single threaded, having to escape 7M keys and 2M values

Execution time on my laptop without this merge request: ~7.2sec vs. with merge request: ~1.8sec.

	@org.junit.Test
	public void lineProtocolPerformanceTest() {

		String[] value = new String[2];
		final long start = System.currentTimeMillis();
		IntStream.range(1, 1000000).forEach((i) -> {
			Map<String, Object> fields1 = Collections.singletonMap("foo ", "bar\""+i);
			Map<String, Object> fields2 = Collections.singletonMap("foo,"+i, "baz\"");

			String measurement = "measurement"+i;

			TimeUnit precision = TimeUnit.NANOSECONDS;

			Map<String, String> tags = Collections.singletonMap("bar \"\\"+i, "b\"a\"z"+i);

			Long time = System.currentTimeMillis();

			Point p1 = new Point();
			p1.setFields(fields1);
			p1.setMeasurement(measurement);
			p1.setPrecision(precision);
			p1.setTags(tags);
			p1.setTime(time);

			Point p2 = new Point();
			p2.setFields(fields2);
			p2.setMeasurement(measurement);
			p2.setPrecision(precision);
			p2.setTags(tags);
			p2.setTime(time);

			value[0] = p1.lineProtocol();
			value[1] = p2.lineProtocol();
		});
		System.out.println(System.currentTimeMillis() - start);
		System.out.println(value[0]);
		System.out.println(value[1]);
	}

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -182,7 +176,7 @@ public Builder fields(final Map<String, Object> fieldsToAdd) {
     /**
      * Add a time to this point.
      *
-     * @param timeToSet the time for this point
+     * @param timeToSet      the time for this point
```
majst01: reset this formatting

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -205,8 +199,8 @@ public Point build() {
       point.setFields(this.fields);
       point.setMeasurement(this.measurement);
       if (this.time != null) {
-          point.setTime(this.time);
-          point.setPrecision(this.precision);
+        point.setTime(this.time);
```
majst01: same here

PR #429, opened 2018-03-15
Parameter binding in InfluxQL (influxdata/influxdb-java#274)

hampelratte:
This is a simple implementation for "prepared statements" (influxdata/influxdb-java#274) using regular expressions.

I added a new query class BoundParameterQuery, which can be used for
"prepared statements". The constructor accepts an InfluxQL expression
with placeholders, a DB name and a varags list to bind the parameters to
the placeholders. I also extended the InfluxDBService, so that the HTTP
requests contain the "params" parameter. The InfluxDBImpl now
differentiates between a Query and a BoundParameterQuery. This is not
the cleanest solution, because this had to be done at a few locations,
but I didn't want to change too much of the code.

It has limitations, because it does not use a parser/grammar for InfluxQL. So you could for example have a placeholder in a string value of a field/tag and this implementation would fail, but at least it is a start and better than risking injections.

Feel free to delete this pull request, if you don't like it at all ;)

### DIFF
--- a/src/main/java/org/influxdb/dto/BoundParameterQuery.java
+++ b/src/main/java/org/influxdb/dto/BoundParameterQuery.java
@@ -12,111 +12,114 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import okio.Buffer;
 
-public class BoundParameterQuery extends Query {
+public final class BoundParameterQuery extends Query {
 
-    private final Map<String, Object> params = new HashMap<>();
+  private final Map<String, Object> params = new HashMap<>();
 
-    private BoundParameterQuery(final String command, final String database) {
-        super(command, database, true);
+  private BoundParameterQuery(final String command, final String database) {
+    super(command, database, true);
+  }
+
+  public String getParameterJsonWithUrlEncoded() {
+    try {
+      List<String> placeholders = parsePlaceHolders(getCommand());
+      assurePlaceholdersAreBound(placeholders, params);
+      String jsonParameterObject = createJsonObject(params);
+      String urlEncodedJsonParameterObject = encode(jsonParameterObject);
+      return urlEncodedJsonParameterObject;
+    } catch (IOException e) {
+      throw new RuntimeException("Couldn't create parameter JSON object", e);
+    }
+  }
+
+  private void assurePlaceholdersAreBound(final List<String> placeholders, final Map<String, Object> params) {
+    if (placeholders.size() != params.size()) {
+      throw new RuntimeException("Unbalanced amount of placeholders and parameters");
     }
 
-    public String getParameterJsonWithUrlEncoded() {
-        try {
-            List<String> placeholders = parsePlaceHolders(getCommand());
-            assurePlaceholdersAreBound(placeholders, params);
-            String jsonParameterObject = createJsonObject(params);
-            String urlEncodedJsonParameterObject = encode(jsonParameterObject);
-            return urlEncodedJsonParameterObject;
-        } catch (IOException e) {
-            throw new RuntimeException("Couldn't create parameter JSON object", e);
-        }
+    for (String placeholder : placeholders) {
+      if (params.get(placeholder) == null) {
+        throw new RuntimeException("Placeholder $" + placeholder + " is not bound");
+      }
+    }
+  }
+
+  private String createJsonObject(final Map<String, Object> parameterMap) throws IOException {
+    Buffer b = new Buffer();
+    JsonWriter writer = JsonWriter.of(b);
+    writer.beginObject();
+    for (Entry<String, Object> pair : parameterMap.entrySet()) {
+      String name = pair.getKey();
+      Object value = pair.getValue();
+      if (value instanceof Number) {
+        writer.name(name).value((Number) value);
+      } else if (value instanceof String) {
+        writer.name(name).value((String) value);
+      } else if (value instanceof Boolean) {
+        writer.name(name).value((Boolean) value);
+      } else {
+        writer.name(name).value(value.toString());
+      }
+    }
+    writer.endObject();
+    return b.readString(Charset.forName("utf-8"));
+  }
+
+  private List<String> parsePlaceHolders(final String command) {
+    List<String> placeHolders = new ArrayList<>();
+    Pattern p = Pattern.compile("\\s+\\$(\\w+?)(?:\\s|$)");
+    Matcher m = p.matcher(getCommand());
+    while (m.find()) {
+      placeHolders.add(m.group(1));
+    }
+    return placeHolders;
+  }
+
+  @Override
+  public int hashCode() {
+    final int prime = 31;
+    int result = super.hashCode();
+    result = prime * result + params.hashCode();
+    return result;
+  }
+
+  @Override
+  public boolean equals(final Object obj) {
+    if (this == obj) {
+      return true;
+    }
+    if (!super.equals(obj)) {
+      return false;
+    }
+    BoundParameterQuery other = (BoundParameterQuery) obj;
+    if (!params.equals(other.params)) {
+      return false;
+    }
+    return true;
+  }
+
+  public static class QueryBuilder {
+    private BoundParameterQuery query;
+    private String influxQL;
+
+    public static QueryBuilder newQuery(final String influxQL) {
+      QueryBuilder instance = new QueryBuilder();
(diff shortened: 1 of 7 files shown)


majst01:
Hi @hampelratte thanks for the effort, we always appreciate contribution.

I think this is a good starting Point, but i would like to have a fluent API without a very wide Constructor. I can think of something like:

```java
query = BoundParameterQuery("SELECT * FROM abc WHERE a > $a AND b < $b")
                  .with("a",0)
                  .with("b",10)
                  .create()
```

src/test/java/org/influxdb/dto/BoundParameterQueryTest.java
```diff
@@ -0,0 +1,188 @@
...
+          .create();
+      query.getParameterJsonWithUrlEncoded();
+      Assert.fail("Expected RuntimeException because of unbound placeholder");
+    } catch (RuntimeException rte) {
+      // expected
+    }
+  }
+
+  @Test
+  public void testEqualsAndHashCode() {
```
fmachado: Is this really necessary? Did you create your own hashcode/equals implementation or you are using the one provided by your IDE?

src/test/java/org/influxdb/dto/BoundParameterQueryTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+  @Test
+  public void testSingleCharacterPlaceHolderParsing() throws IOException {
+    BoundParameterQuery query = QueryBuilder.newQuery("SELECT * FROM abc WHERE a > $a AND b < $b")
+        .forDatabase("foobar")
+        .bind("a", 0)
+        .bind("b", 10)
+        .create();
+    Map<String, Object> params = readObject(decode(query.getParameterJsonWithUrlEncoded()));
+    Assert.assertEquals(2, params.size());
```
fmachado: I really appreciate your detailed and extensive test but I think it's too much: you are testing multiple times the same logic over and over.

Could you please test `query.getParameterJsonWithUrlEncoded()` only once, covering all data types supported by InfluxDB?

src/main/java/org/influxdb/dto/BoundParameterQuery.java
```diff
@@ -0,0 +1,125 @@
...
+
+  private final Map<String, Object> params = new HashMap<>();
+
+  private BoundParameterQuery(final String command, final String database) {
+    super(command, database, true);
+  }
+
+  public String getParameterJsonWithUrlEncoded() {
+    try {
+      List<String> placeholders = parsePlaceHolders(getCommand());
```
fmachado: `parsePlaceHolders` is only used to create a list for `assurePlaceholdersAreBound`.
Could you please move `parsePlaceHolders` to be called from inside `assurePlaceholdersAreBound`?

src/main/java/org/influxdb/dto/BoundParameterQuery.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  private void assurePlaceholdersAreBound(final List<String> placeholders, final Map<String, Object> params) {
+    if (placeholders.size() != params.size()) {
+      throw new RuntimeException("Unbalanced amount of placeholders and parameters");
+    }
+
+    for (String placeholder : placeholders) {
+      if (params.get(placeholder) == null) {
+        throw new RuntimeException("Placeholder $" + placeholder + " is not bound");
```
fmachado: Personally I don't like to throw `RuntimeException`s (check why here https://wiki.sei.cmu.edu/confluence/display/java/ERR07-J.+Do+not+throw+RuntimeException%2C+Exception%2C+or+Throwable ).

Also, this exception will be thrown only when the query is about to be executed instead of the moment when I'm executing the `.create()` from the Builder. Would be possible to move it to the builder?

Another question: what about throwing `IllegalStateException` from the `.create()` method?

fmachado:
@hampelratte and before I forget, thanks for your contribution. :)

hampelratte:
@fmachado Good point. I was thinking, that a check for the placeholders might be a good thing, but obviously InfluxDB does that already and returns reasonable error messages. So there is no reason to do that in the connector. I'm going to remove that and as a result the majority of your annotations will get resolved, too.

src/main/java/org/influxdb/dto/BoundParameterQuery.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  private String createJsonObject(final Map<String, Object> parameterMap) throws IOException {
+    Buffer b = new Buffer();
+    JsonWriter writer = JsonWriter.of(b);
+    writer.beginObject();
+    for (Entry<String, Object> pair : parameterMap.entrySet()) {
+      String name = pair.getKey();
+      Object value = pair.getValue();
+      if (value instanceof Number) {
```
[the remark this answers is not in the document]
hampelratte: I had a look at the moshi code and tested it. This works as intended.

src/main/java/org/influxdb/dto/BoundParameterQuery.java
```diff
@@ -0,0 +1,125 @@
...
+        writer.name(name).value((Number) value);
+      } else if (value instanceof String) {
+        writer.name(name).value((String) value);
+      } else if (value instanceof Boolean) {
+        writer.name(name).value((Boolean) value);
+      } else {
+        writer.name(name).value(value.toString());
```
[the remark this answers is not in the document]
hampelratte: Done

src/test/java/org/influxdb/dto/BoundParameterQueryTest.java
hampelratte: There is a test for hashcode and equals in Query. So I just did it for BoundParameterQuery, too.

src/main/java/org/influxdb/dto/BoundParameterQuery.java
```diff
@@ -0,0 +1,125 @@
...
+      String jsonParameterObject = createJsonObject(params);
+      String urlEncodedJsonParameterObject = encode(jsonParameterObject);
+      return urlEncodedJsonParameterObject;
+    } catch (IOException e) {
+      throw new RuntimeException("Couldn't create parameter JSON object", e);
+    }
+  }
+
+  private void assurePlaceholdersAreBound(final List<String> placeholders, final Map<String, Object> params) {
+    if (placeholders.size() != params.size()) {
```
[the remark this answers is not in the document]
hampelratte: Relevant code has been removed

src/main/java/org/influxdb/dto/BoundParameterQuery.java
```diff
@@ -0,0 +1,125 @@
...
+        writer.name(name).value(value.toString());
+      }
+    }
+    writer.endObject();
+    return b.readString(Charset.forName("utf-8"));
+  }
+
+  private List<String> parsePlaceHolders(final String command) {
+    List<String> placeHolders = new ArrayList<>();
+    Pattern p = Pattern.compile("\\s+\\$(\\w+?)(?:\\s|$)");
```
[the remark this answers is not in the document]
hampelratte: Relevant code has been removed

src/test/java/org/influxdb/dto/BoundParameterQueryTest.java
hampelratte: Relevant code has been removed

src/main/java/org/influxdb/dto/BoundParameterQuery.java
hampelratte: Relevant code has been removed

src/main/java/org/influxdb/dto/BoundParameterQuery.java
hampelratte: Relevant code has been removed

fmachado:
@majst01 do you want to take a look?

fmachado:
@hampelratte could you please update README.md with an explanation and example of this new feature? Also, would be nice to have a link to the this PR on CHANGELOG.md under a new section *"2.10 [unreleased]"*.

majst01:
Hi,
yes, once we have a bit documentation and a changelogentry i will merge this.
Again, thanks @hampelratte  for contribution !

majst01:
Hi,
Readme.md ist fixed.
If you are using maven or gradle to build and package your application, then okio is part of the transitive dependencies, if not you must of course care by yourself.

PR #432, opened 2018-03-20
Fixed setAccessible modif. on POJOs and changes are not reversed anymore

### LINKED ISSUE #430: InfluxDBResultMapper not thread safe?
I am refering to this part of the documentation:

InfluxDBResultMapper resultMapper = new InfluxDBResultMapper(); // thread-safe - can be reused
List<Cpu> cpuList = resultMapper.toPOJO(queryResult, Cpu.class);

The problem is when I perform a multi-threaded mapping similar to this part of the code in several threads in parallel:

List<Cpu> cpuList = resultMapper.toPOJO(queryResult, Cpu.class);

I do get the following exception:

org.influxdb.InfluxDBMapperException: java.lang.IllegalAccessException: Class org.influxdb.impl.InfluxDBResultMapper can not access a member of class org.influx.test.Cpu with modifiers "private"
        at org.influxdb.impl.InfluxDBResultMapper.parseSeriesAs(InfluxDBResultMapper.java:184)
        at org.influxdb.impl.InfluxDBResultMapper.lambda$null$2(InfluxDBResultMapper.java:100)

The problem is most likely that the first thread did already set the accessibility of a field back to its original value when the second thread tries to write the field.

fmachado:
Fix #430

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
@@ -203,9 +203,10 @@ public class InfluxDBResultMapper {
       return;
     }
     Class<?> fieldType = field.getType();
-    boolean oldAccessibleState = field.isAccessible();
     try {
-      field.setAccessible(true);
+      if (!field.isAccessible()) {
+        field.setAccessible(true);
+      }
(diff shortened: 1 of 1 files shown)


majst01:
@fmachado you can merge by yourself, i cant comment on this.

PR #434, opened 2018-03-23
allow to figure out, whether the Point.Builder has any fields

kub:
since the build mehtod contains validation for fields emptiness, there should be also way, how to figure out, whether the Builder contains any fields (to prevent the build method from throwing an exception)

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -188,6 +188,15 @@ public class Point {
     }
 
     /**
+     * Does this builder contain any fields?
+     *
+     * @return true, if the builder contains any fields, false otherwise.
+     */
+    public boolean hasFields() {
+      return !fields.isEmpty();
+    }
+
+    /**
      * Create a new Point.
      *
      * @return the newly created Point.
--- a/src/test/java/org/influxdb/dto/PointTest.java
+++ b/src/test/java/org/influxdb/dto/PointTest.java
@@ -330,4 +330,13 @@ public class PointTest {
 		// THEN equals returns true
 		assertThat(equals).isEqualTo(false);
 	}
+
+	@Test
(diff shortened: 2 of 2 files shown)


majst01:
What exact problem are you trying to solve, can you show it in a testcase ? Dont see the point actually.

kub:
I'll try to explain my problem more into the detail, hope it helps: 
as the ```Point``` is constructed by the ```Point.Builder```,  ```Point.Builder#build()``` can be called in different part of the application (different method), than ```Point.Builder#addField```. In such case, there is no way how to avoid ```IllegalArgumentException```, that is thrown from [here](https://github.com/influxdata/influxdb-java/blob/master/src/main/java/org/influxdb/dto/Point.java#L197) in case, that no fields were actually added (e.g. because data of these fields was null). 

It's also convenient for other cases. Imagine e.g. this one:
```Groovy
void reportValues(String value1, String value2, String value3) {
    def builder = Point.measurement(measurement).tag(tags).time(time)

    if(value1 != null) {
        builder.addField(TIMESERIES_1, value1)
    }

    if(value2 != null) {
        builder.addField(TIMESERIES_2, value2)
    }

    if(value3 != null) {
        builder.addField(TIMESERIES_3, value3)
    }

    builder.build() // throws IllegalArgumentException, if all values are null

}
```
... again - there should be a way, how to 'ask' the builder, whether it is ok, to call the ```build``` method or whether it will fail with the ```IllegalArgumentException```.

majst01:
Ok understood, but then you need to add a testcase and a addition to the usage description in the README.md to you MR.

kub:
I've added unit test

I'm not sure about README.md - do you really think, that this small detail should be really added there? Tried to read through it and honestly - I'm not sure where to put it. Also seems to me, that README.md should not go into that level of detail (just for sake of readability) - what do you think?

PR #435, opened 2018-03-24
Speedup travis build

majst01:
Build should always be as fast as possible. so i did:

- create a travis matrix of JDK and Influxdb version
- run test against all influxdb versions only with the most current JDK, test with JDK8 only influxdb 1.5
- reduce the amount of points written in one particular Performancetest.

Total duration is now ~3min compared to 35min before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,19 @@
+language: java
+
 sudo: required
 
 services:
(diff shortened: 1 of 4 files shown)


majst01:
@fmachado what do you think ?

Issue #413, opened 2018-01-19
Debug mode which allows HTTP requests being sent to the database to be logged


lxhoan:
It could be more convenient, for the java client, if we  have a debug mode which allows HTTP requests being sent to the database to be logged

majst01:
If you call:
```java
influxDB.setLogLevel(LogLevel.FULL) 
```
You will get all http payload including headers in the logs.

ivankudibal:
I'd keep this open.. The problem is the operations people have no tools to set the full logging.. eg via external property

fmachado:
Ah, now I got it. Would be OK for you to expose logging control and diagnostic related data via JMX? Do you have something else in mind?

ivankudibal:
1/ System property is should and satisfactory (in 2.9) and then 2/ JMX
would be really nice to have

majst01:
Is this required to implement here, we provide a method to set/change the loglevel, if a application want to make these changes dynamically it can be done in one of the ways you mentioned, i would like to keep this implementation small.

lxhoan:
I updated the issue description, add acceptance criterion

PR #450, opened 2018-05-24
issues #413 : Debug mode which allows HTTP requests being sent to the…

lxhoan:
… database to be logged


src/test/java/org/influxdb/InfluxDBLogLevelTest.java
```diff
@@ -0,0 +1,50 @@
...
+public class InfluxDBLogLevelTest {
+  @Test
+  public void testLogLevelProperties() {
+    
+    String oldLogLevel = System.getProperty(InfluxDB.LOG_LEVEL_PROP);
+    List<String> logLevels = Arrays.asList(null, "NONE", "BASIC", "HEADERS", "FULL", "abc");
+    logLevels.forEach(logLevel -> {
+      System.out.println("LogLevel = " + logLevel);
+      Optional.ofNullable(logLevel).ifPresent(value -> {
+        System.setProperty(InfluxDB.LOG_LEVEL_PROP, value);
```
fmachado: I don't think it's a good idea to change global configurations (in this case, defining global properties) during the test as this **may** affect other tests if they are running in parallel. Also, if your test fail here for some reason, you will exit the test case (line 38: fail(e)) and **may** leave the system with an incorrect level configuration.
My example may be too much for this simple test but my point here is to make you aware that this is, in general, a bad practice.

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -739,4 +739,28 @@ public void dropRetentionPolicy(final String rpName, final String database) {
     execute(this.influxDBService.postQuery(this.username, this.password,
         Query.encode(queryBuilder.toString())));
   }
+
+  private void setLogLevel() {
```
fmachado: - This method is being called every time an object is created. Can we call it just once, statically?
- IMO the name of this method is not following the JavaBean convention: `setXXX` usually is used to change the state and expects 1..* parameter(s). Can you rename it to something like `configureLogLevel`?
- I'm almost sure all this method can be replaced with something like:
```Java
String value = System.getProperty...
if (value != null) {
   // logLevel = ... don't change this property with this method as there is already another method changing it.
   // instead, call directly setLogLevel(...)
   setLogLevel(LogLevel.valueOf(value.toUpperCase()));
}
```

lxhoan: >This method is being called every time an object is created. Can we call it just once, statically?

I dont think so, I think we should allow developer to change the property value on the fly (as you can see in the unit test) for debugging purpose, or its visibility may differ from context to context (custom SecurityManager is used)

>IMO the name of this method is not following the JavaBean convention: setXXX usually is used to change the state and expects 1..* parameter(s). Can you rename it to something like configureLogLevel?

For me the naming convention setXXX(...) and setXXX() are not semantically ambiguous. But rename it to initLogLevel anyway

>I'm almost sure all this method can be replaced with something like

partially agree and fix accordingly, except keeping switch case since you still have check for null or invalid property value (either by conditional structure like this or catching exception from valueOf(String))

src/test/java/org/influxdb/InfluxDBLogLevelTest.java
lxhoan: add try/finally block

For now I see the Surefire plugin config runs tests in sequential. however add NotThreadSafe annotation to make sure Surefire will execute the test sequentially
http://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html

src/main/java/org/influxdb/impl/InfluxDBImpl.java
fmachado: > I dont think so, I think we should allow developer to change the property value on the fly (as you can see in the unit test) for debugging purpose
So, you want to provide a way to the developer to change the loglevel of InfluxDB objects already instantiated in order to make the debug easier. Is this correct? If so, then you expect the developer to create on his side something like what you did on the test case: `System.setProperty`. How do you expect that the developer will change the loglevel while the application is running?
If I got your intentions correctly and what I said before is right then I suggest you to do the right way and use JMX.

fmachado: > For me the naming convention setXXX(...) and setXXX() are not semantically ambiguous. But rename it to initLogLevel anyway

Please check the Section "8.3 Design Patterns for Properties" from the JavaBean API specification: http://download.oracle.com/otn-pub/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/beans.101.pdf .

fmachado: > partially agree and fix accordingly, except keeping switch case since you still have check for null or invalid property value (either by conditional structure like this or catching exception from valueOf(String))
Ok., I don't think you got my point of view. Let me be more clear: If you have 100 different loglevels, would you still keep using a switch/case statement? AFAIK, your code can be replaced and "secured" with a try/catch with no collateral effects.

src/test/java/org/influxdb/InfluxDBLogLevelTest.java
```diff
@@ -0,0 +1,53 @@
...
+import net.jcip.annotations.NotThreadSafe;
+
+/**
+ * Test the InfluxDBimpl log level setting from system property.
+ *
+ * @author hoan.le [at] bonitoo.io
+ *
+ */
+@RunWith(JUnitPlatform.class)
+@NotThreadSafe
```
fmachado: Are you 100% sure this annotation will affect how JUnit works?
Anyway, instead of fixing the bad practice you introduced by making changes that may affect other tests you are here trying to do something that would make the tests not going in parallel? I don't get it.

src/main/java/org/influxdb/impl/InfluxDBImpl.java
lxhoan: >So, you want to provide a way to the developer to change the loglevel of InfluxDB objects already instantiated in order to make the debug easier. Is this correct? If so, then you expect the developer to create on his side something like what you did on the test case: System.setProperty. How do you expect that the developer will change the loglevel while the application is running?
If I got your intentions correctly and what I said before is right then I suggest you to do the right way and use JMX.

As far as I know JMX is a technology for managing and monitoring rather than debugging the application. For debugging you would like to use a more flexible but simple way, and use of a System property here would be better. Two use cases:
1/ You can pass the log level to JVM at booting up
2/ During your debugging session, you can manage to set the log level of the client which is going to created somewhere else (I do not reject you can use JMX but it will never have such a flexibility like evaluating a java statement to change that System property value)

lxhoan: >Please check the Section "8.3 Design Patterns for Properties" from the JavaBean API specification: http://download.oracle.com/otn-pub/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/beans.101.pdf .

I've had this already, so initLogLevel() is ok for you or not ?

lxhoan: >Ok., I don't think you got my point of view. Let me be more clear: If you have 100 different loglevels, would you still keep using a switch/case statement? AFAIK, your code can be replaced and "secured" with a try/catch with no collateral effects.

OK, got it, fixed as your comment (use of Enum.valueOf(,))
Tks

src/test/java/org/influxdb/InfluxDBLogLevelTest.java
dubsky: Hi,

I went through the code with the following findings:

1) InfluxDB.LOG_LEVEL_PROP ? It should be InfluxDB.LOG_LEVEL_PROPERTY. It is 5 chars more only.
2) I don't even like the 'initLogLevel' construct. Rather we should do:

   `  this.setLogLevel(this.getConfiguredLogLevel())`

getConfiguredLogLevel() would just read the configuration property and set a default value for the LogLevel. Like this, we have two functions where it is clear what is happening here and there this one function less that has side effects.

Vlada

dubsky:
Just repeating myself:

I went through the code with the following findings:

InfluxDB.LOG_LEVEL_PROP ? It should be InfluxDB.LOG_LEVEL_PROPERTY. It is 5 chars more only.

I don't even like the 'initLogLevel' construct. Rather we should do:

this.setLogLevel(this.getConfiguredLogLevel())

getConfiguredLogLevel() would just read the configuration property and set a default value for the LogLevel. Like this, we have two functions where it is clear what is happening here and there this one function less that has side effects.
3) If the testing is a problem we may even do (simplified without error processing):

this.setLogLevel(this.getConfiguredLogLevel(System.getProperty(InfluxDB.LOG_LEVEL_PROPERTY)))

In this case, we may test the getConfiguredLogLevel function individually. (If we agree that we don't need to test System.getProperty) And we would avoid the discussion about forcing sequential test execution.

Vlada

src/test/java/org/influxdb/InfluxDBLogLevelTest.java
lxhoan: @dubsky , thank you Vlada,
I will fix as @fmachado requested, for me it's not too bad

fmachado:
@dubsky fine by me and just to be clear: my point was not about testing or not something specific. It was originally about changing state(s) that are globally visible and may or may not affect concurrent tests.

Something else that I mentioned is regarding changing some configurations at runtime by calling `System.setProperty`. Now I promise this will be my last arguments.

Yes, I agree that will work in some cases but it's not standardized and the developer will have to find a way to trigger this `System.setProperty` somehow. Let me give you some concrete examples:
1) On projects that I'm responsible for, we don't expose maintenance operations via REST services and this is a security measure. How am I supposed to trigger this `System.setProperty`?
2) I have a few JavaSE apps running (they are doing ML). How am I supposed to trigger this `System.setProperty` without stopping the apps?

I can solve all previous issues using JMX but not with `System.setProperty`. By the way, I'm not inventing this approach. With a quick google search, I could find a few libraries[1][2][3][4] using JMX to expose metrics and/or change configuration.

[1] ehcache http://www.ehcache.org/documentation/2.7/operations/jmx.html
[2] logback https://logback.qos.ch/manual/jmxConfig.html
[3] log4j https://logging.apache.org/log4j/2.0/manual/jmx.html
[4] cassandra java driver https://docs.datastax.com/en/drivers/java/3.3/com/datastax/driver/core/Metrics.html

@dubsky @majst01 @ivankudibal what do you think?

lxhoan:
Just want to quote here this (from a discussion between @fmachado and me)

>As far as I know JMX is a technology for managing and monitoring rather than debugging the application. For debugging you would like to use a more flexible but simple way, and use of a System property here would be better. Two use cases:
1/ You can pass the log level to JVM at booting up
2/ During your debugging session, you can manage to set the log level of the client which is going to created somewhere else (I do not reject you can use JMX but it will never have such a flexibility like evaluating a java statement to change that System property value)

So unless we change the original issue to something like __Operation management and monitoring mechanism for influxdb-java ... __ , I would keep on with System property. There are thousands of real practices we can find easily on the Internet for using of System property to configure certain behaviour(s) of an application. Hence, I think I do not need to list some of them here.

However I agree with that I will remove the Use Case /2. So in case if this system property exists, any change to it's value  will not take effect and we can set the logLevel in a static context which is executed only one time. I hope this change is aligned with the review

>This method is being called every time an object is created. Can we call it just once, statically?

lxhoan:
@dubsky , @fmachado 

I updated the issue #413 to clarify the objectives of this PR
I fix the implementation with regard to the updated issue and your comments, please help to review.

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -59,6 +59,7 @@
   static final okhttp3.MediaType MEDIA_TYPE_STRING = MediaType.parse("text/plain");
 
   private static final String SHOW_DATABASE_COMMAND_ENCODED = Query.encode("SHOW DATABASES");
+  private static final String DEBUG_MODE_LOG_LEVEL = System.getProperty(LOG_LEVEL_PROPERTY);
```
majst01: small nit, why DEBUG_MODE_LOG_LEVEL ? would not LOG_LEVEL be more accurate ?

majst01:
I am fine with this, my single comment regarding property naming might be of interest, but whats really missing is a small hint in the documentation and a CHANGELOG.md entry. Thanks

lxhoan:
@majst01 , I added change log and source documentation, thank you

PR #453, opened 2018-06-05
add method to map result of GROUP BY query

MrBeepoint:
The added method toPOJOMap adds functionality to convert a QueryResult from a GROUP BY query to a map of lists containing the result entries. The keys of the map are the concatenated tag values in the QueryResult.

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
@@ -27,8 +27,11 @@ import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Objects;
+import java.util.Map;
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


fmachado:
@MrBeepoint first of all, thanks for the contribution!
I'm almost sure you can achieve the same goal using Streams without having to change the influxdb-java source code and perform string concatenation on tags.
Could you please take a look on http://www.technicalkeeda.com/java-8-tutorials/java-8-stream-grouping ?.

MrBeepoint:
@fmachado Thanks for the hint regarding the grouping feature of streams. I guess this could be a possible solution for my application, because I do not have tons of data. But by using a stream and group data after already having the list of data objects, the same number of data objects have to be processed again. Maybe this could be a problem for a huge amount of data objects.

I am not very familiar with streams and their performance. So do you think it makes sense to do a comparison (benchmark) between both implementations?

fmachado:
@MrBeepoint were you able to use Streams with your solution? If so then would be great f you could add one or two examples to the main documentation. What do you think? :)

PR #454, opened 2018-06-07
Adding wait time for writing a point to the DB

sankalpanand:
We've been facing a problem in production for a pretty widely used product. `BatchProcessor` is currently using a BlockingQueue. There have been times when writing to this queue has blocked our main thread while writing a metric into the queue. 

I'm trying to address that issue through this PR by using `offer()` with a timeout instead of `put`. This will allow us to simply drop that data point if its taking too long instead of blocking the main thread.


src/main/java/org/influxdb/impl/BatchProcessor.java
```diff
@@ -354,6 +354,22 @@ public void run() {
     }
   }
 
+  void put(final AbstractBatchEntry batchEntry, final long timeout) {
+    try {
+      this.queue.offer(batchEntry, timeout, TimeUnit.SECONDS);
```
efesenmaier: Should be in `TimeUnit.MILLISECONDS`

PR #458, opened 2018-06-09
Adding wait time for writing a point to the DB

sankalpanand:
We've been facing a problem in production for a pretty widely used product. BatchProcessor is currently using a BlockingQueue. There have been times when writing to this queue has blocked our main thread while writing a metric into the queue.

I'm trying to address that issue through this PR by using offer() with a timeout instead of put. This will allow us to simply drop that data point if its taking too long instead of blocking the main thread.

### DIFF
--- a/src/main/java/org/influxdb/InfluxDB.java
+++ b/src/main/java/org/influxdb/InfluxDB.java
@@ -284,6 +284,20 @@ public interface InfluxDB {
   public void write(final String database, final String retentionPolicy, final Point point);
 
   /**
+   * Write a single Point to the database by waiting for specified time for space to become available.
+   *
+   * @param database
+   *            the database to write to.
+   * @param retentionPolicy
+   *            the retentionPolicy to use.
+   * @param point
+   *            The point to write
+   * @param timeout
+   *            How long to wait in seconds
+   */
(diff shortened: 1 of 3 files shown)


majst01:
Hi,
first of all thank you for your contribution, problem understood, but i would like to discuss this a bit more. 

- I really would like to prevent to wider the existing interface any more, it is already to complicated and therefore harder to use correctly
- if your type of usage scenario is valid, i assume actually, we should think about if we cannot achieve this with more intuitive interface.
- Without having thought to deep, wouldnt that also possible on BatchProcessor creation ?
- When using ```offer``` a boolean is returned which indicates the operation succeeded or not but you dont handle this. 

And any PR´s submitted must have tests included which are able to fully cover the code path added and enlarge the coverage overall.

efesenmaier:
Is there any problem using the unbounded version of the LinkedBlockingQueue in all cases? That solves the problem of blocking on write. The batch size may vary some from batch to batch, but I do not think that will cause any harm. In that case, the "batch size" is simply the queue size in which a write will be triggered, but the queue is technically unbounded.

majst01:
If we use a unbounded LinkedBlockingQueue we could potentially come to a OOM if writing to the Batchpoints is way faster than flushing happens.

majst01:
stale, no tests, closing

Issue #389, opened 2017-12-01
Support for MessagePack


dvanwest:
Is it possible to update the InfluxDB Java API to allow query data to be returned in MessagePack format?  This format is supported. when reading, in the new 1.4 version of InfluxDB.

majst01:
we use retrofit for the backend communication and therefore a converter for messagepack is needed.
Luckily such a converter exists:

https://github.com/komamitsu/retrofit-converter-msgpack

PR #460, opened 2018-06-13
Implement Issue #389 : Support for MessagePack

lxhoan:
thi PR for Issue #389 : Support for MessagePack
I use the https://github.com/komamitsu/retrofit-converter-msgpack (thanks to @majst01)

### DIFF
--- a/src/main/java/org/influxdb/dto/QueryResult.java
+++ b/src/main/java/org/influxdb/dto/QueryResult.java
@@ -3,6 +3,8 @@ package org.influxdb.dto;
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
 /**
  * {Purpose of This Type}.
  *
@@ -68,9 +70,19 @@ public class QueryResult {
     this.error = error;
   }
 
+  @JsonIgnoreProperties(ignoreUnknown = true)
   public static class Result {
+    private List<Abc> abc;
     private List<Series> series;
     private String error;
+
+    public List<Abc> getAbc() {
+      return abc;
+    }
+
+    public void setAbc(final List<Abc> abc) {
+      this.abc = abc;
+    }
 
     /**
      * @return the series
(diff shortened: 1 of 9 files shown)


src/main/java/org/influxdb/dto/QueryResult.java
```diff
@@ -3,6 +3,8 @@
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
```
majst01: Why do we introduce a new dependency to jackson ?

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -1,12 +1,18 @@
 package org.influxdb.impl;
 
 
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.MappingIterator;
+import com.fasterxml.jackson.databind.MappingJsonFactory;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
```
majst01: Same here, i really would like to prevent to switch from moshi to jackson with this PR. I am not against changing the json converter library, i just want to do it by "accident" when introducing messagepack format. We should do this move, once decided intentionally and without any other modifications.

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -99,34 +124,94 @@ public InfluxDBImpl(final String url, final String username, final String passwo
...
+    chunkProccesor = (chunkedBody, consumer) -> {
+        BufferedSource source = chunkedBody.source();
+
+        QueryResult result = null;
+        MappingIterator<QueryResult> results = objectReader.readValues(source.inputStream());
+        while (results.hasNext()) {
+          result = results.nextValue();
+          consumer.accept(result);
+        }
+    };
```
majst01: Way too much code in the constructor.

src/test/java/org/influxdb/BatchOptionsTest.java
```diff
@@ -130,7 +132,8 @@ public void testFlushDuration() throws InterruptedException {
       
       //check no points writen to DB before the flush duration
       QueryResult result = influxDB.query(new Query("select * from weather", dbName));
-      Assertions.assertNull(result.getResults().get(0).getSeries());
+      List<Series> series = result.getResults().get(0).getSeries();
```
majst01: Any reason why a test must be changed ?

src/main/java/org/influxdb/impl/InfluxDBImpl.java
fmachado: @majst01 I believe Jackson is being used here because it's a direct dependency used by the msgpack converter library, as you can see [here](https://github.com/komamitsu/retrofit-converter-msgpack/blob/master/build.gradle#L36).

src/main/java/org/influxdb/impl/InfluxDBImpl.java
lxhoan: In a minor release, I would like to keep the implementation as back-ward compatible as possible. this constructor

```
  InfluxDBImpl(final String url, final String username, final String password, final OkHttpClient.Builder client,
          final InfluxDBService influxDBService, final JsonAdapter<QueryResult> adapter) {
```
in my opinion, even it seems awkward and used nowhere in the source code, we are not used if it is used by any user or not so I decided to keep it. That lead me to introduce the ChunkProcessor so we can have polymorphism for processing __chunking query__

@majst01, I believe Jackson has a better API on processing HTTP chunked response

src/test/java/org/influxdb/BatchOptionsTest.java
lxhoan: Because:
   +in JSON response (for an empty query), the 'series' element does not exists
   + in MessagePack response  (for an empty query), series exist as an empty array
--> I think none of them are not wrong, just the consistency from server side. I am actually thinking to open an issue for the server

src/main/java/org/influxdb/impl/InfluxDBImpl.java
lxhoan: again, the convertor depends on Jackson

src/main/java/org/influxdb/dto/QueryResult.java
lxhoan: because the converter https://github.com/komamitsu/retrofit-converter-msgpack (you proppsed it) depends on jackson

majst01: OK, i didnt check that, my fault. But still, using another json converter with this change is not desired. We must think twice "

lxhoan:
closed because PR #471 is the final PR

PR #474, opened 2018-07-23
Remove code which checks for unsupported influxdb versions

majst01:


lxhoan:
Hi @majst01 , 
I'd like to ask if this PR means support for versions 0.x and 1.0x are now dropped ? If so I agree that checking for them is not neccessary

majst01:
Yes, the actual version cannot talk to version 1.0 and below because of the changed lineprotocol.

PR #478, opened 2018-07-26
fix issue #445 : https://github.com/influxdata/influxdb-java/issues/445

lxhoan:
Hi @majst01 , I accidentally dropped PR #477 so create this new one
PR is now mergable, please take a look, tks

### DIFF
--- a/src/test/java/org/influxdb/InfluxDBTest.java
+++ b/src/test/java/org/influxdb/InfluxDBTest.java
@@ -18,6 +18,8 @@ import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 
+import okhttp3.OkHttpClient;
+
 import java.io.IOException;
 import java.time.Instant;
 import java.time.ZoneId;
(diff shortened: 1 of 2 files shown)


src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -891,5 +895,63 @@ public void testMessagePackOnOldDbVersion() {
...
+    for (int i = 0; i < maxCallables; i++) {
+      callableList.add(new Callable<String>() {
+        @Override
+        public String call() throws Exception {
+          MyInfluxDBBean myBean = new MyInfluxDBBean();
+          return myBean.connectAndDoNothing1();
+        }
+      });
+    }
+    System.out.println("Invoking all callableList (size()=" + callableList.size() + ")");
```
majst01: I am not a big fan of having sysout in tests, are these necessary ?

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -18,21 +36,7 @@
 import org.junit.platform.runner.JUnitPlatform;
 import org.junit.runner.RunWith;
 
-import java.io.IOException;
```
majst01: please revert this formatting

src/test/java/org/influxdb/InfluxDBTest.java
lxhoan: removed them

src/test/java/org/influxdb/InfluxDBTest.java
lxhoan: reverted as comments, thanks

Issue #480, opened 2018-07-31
UDP target host address is cached


m4ce:
Hi,

the target host address used in the UDP writer appears to be cached on initialization. However, this is problematic if what `url` resolves to changes. We are running our apps on a cloud platform where service discovery is done over DNS. Because of this caching, the application will silently stop sending metrics over UDP as an old ip address will be erroneously used.

fmachado:
This issue was fixed by PR #481.

PR #481, opened 2018-07-31
Dynamic UDP target host address

m4ce:
This PR tries to solve #480 by refreshing the target UDP host address every 5m. I am not sure if this is the best approach but that's all I could come up with!

### DIFF
--- a/src/main/java/org/influxdb/InfluxDBFactory.java
+++ b/src/main/java/org/influxdb/InfluxDBFactory.java
@@ -1,9 +1,8 @@
 package org.influxdb;
 
+import okhttp3.OkHttpClient;
 import org.influxdb.InfluxDB.ResponseFormat;
 import org.influxdb.impl.InfluxDBImpl;
-
-import okhttp3.OkHttpClient;
 import org.influxdb.impl.Preconditions;
 
 import java.util.Objects;
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -45,6 +45,8 @@ import java.net.InetAddress;
 import java.net.SocketException;
 import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Executors;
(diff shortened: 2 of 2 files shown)


lxhoan:
Hi @fmachado , I think resolving and caching IP address for datagram is unnecessary (and obviously falls defective this case)
In my opinion, this snippet 
```
datagramSocket.send(new DatagramPacket(bytes, bytes.length, hostAddress, udpPort));
```
hostAddress should be changed to SocketAdress

What do you think ?

m4ce:
@lxhoan, that works well only if one has batching enabled. If batching is disabled,  it will resolve the hostname every time an UDP write is performed. It can result in denial of service to the DNS service if one sends out a lot of metrics in a short interval of time.

lxhoan:
@m4ce, there some other dns caching levels down to the IP layer (at least at JVM and your OS), So I think we can be safe
(for java you can refer to this https://docs.oracle.com/javase/6/docs/technotes/guides/net/properties.html)

Otherwise we must be meticulous when introducing more configuration to influxdb (I mean this one ADDRESS_LOOKUP_INTERVAL)

lxhoan:
Hi @fmachado , if you can arrange time please take a look on it
I think this is a realistic use case that influxdb-java should consider to support

fmachado:
@m4ce as far as I could understand from your issue ticket and this PR, you can solve your problem by adding a parameter `-Dnetworkaddress.cache.ttl=300L` to the JVM (as mentioned by @lxhoan and [here](https://docs.oracle.com/javase/8/docs/technotes/guides/net/properties.html)).

You can also apply this configuration by calling `Security.setProperty` ([javadoc](https://docs.oracle.com/javase/8/docs/api/java/security/Security.html)) but I don't know what happens with already resolved names.

For the reasons mentioned before, I vote for do not merge this PR.

fmachado:
I'm closing this issue but feel free to reopen it if you think your issue cannot be fixed with the way we mentioned here.

m4ce:
@fmachado , sorry to comment on a close issue. I think you suggestion would only work if we changed the code to be:

```datagramSocket.send(new DatagramPacket(bytes, bytes.length, new InetSocketAddress(hostname,  udpPort)));
```

Just like @lxhoan suggested above.

m4ce:
I see this change was  implemented on Sept 5th. We are good then!

PR #484, opened 2018-08-08
Feature/add precision for batch write mode

lvqian:


lvqian:
this pr aim to fix batch write mode losing time precision issue

lvqian:
Is there any expert can take a look at this PR? thanks in advance.

lxhoan:
Hi @lvqian ,
>this pr aim to fix batch write mode losing time precision issue

BatchPoints.precision set to NANOSECONDS by default.

Could you justify for the reason why you need to use the precision of the first added point as the precision of the whole batch ? (I dont think it's a very common use case)

majst01:
Hi @lvqian 

I also do not exactly understand what you want to achieve, can you please describe in an issue what problem you experience, probably with sample code. 

If this is a performance related PR, we should also see some performance numbers to be sure we get what we want. 
Furthermore adding more visibility to methods which are protected right now is not a good practice, as it widens the actual API. 

And lastly, all code added should be covered by additional tests.

lvqian:
Hi @majst01 

Sorry for the late reply. The issue we faced was, in our actual usage, we only need send second precision points to influxdb, but using current influxdb java sdk, the actually sent point with default nanosecond precision. We want lower the precision to second.

Issue #485, opened 2018-08-13
MessagePack queries: Exception during parsing InfluxDB version [macOS]


bednar:
Hi @lxhoan, 

I want to try the MessagePack queries at **macOS** but there is a bug in parsing InfluxDB version. On the macOS the version of InfluxDB is in a format **v1.6.0** so [parsing of the version](https://github.com/influxdata/influxdb-java/blob/becbfb59b61662b8171bc0dfaf2b343066a9798c/src/main/java/org/influxdb/impl/InfluxDBImpl.java#L672) fails.

The InfluxDB was installed by [brew](Homebrew). 

I also created a [simple test](https://github.com/bonitoo-io/influxdb-java/blob/68afd40561d8799e896936d12b3798eca8a21cb0/src/test/java/org/influxdb/MessagePackInfluxDBTest.java#L190) so you can reproduce it.

### The InfluxDB ping response on macOS
```
HTTP/1.1 204 No Content
Content-Type: application/json
Request-Id: ab0cd803-9ee2-11e8-8d8b-000000000000
X-Influxdb-Build: OSS
X-Influxdb-Version: v1.6.0
X-Request-Id: ab0cd803-9ee2-11e8-8d8b-000000000000
Date: Mon, 13 Aug 2018 10:21:40 GMT
```

### The Exception

```java
java.lang.NumberFormatException: For input string: "v1"

	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
	at org.influxdb.impl.InfluxDBImpl.executeQuery(InfluxDBImpl.java:672)
	at org.influxdb.impl.InfluxDBImpl.query(InfluxDBImpl.java:507)
	at org.influxdb.MessagePackInfluxDBTest.parsingVersionV1_6_0(MessagePackInfluxDBTest.java:201)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:513)
[...]

PR #487, opened 2018-08-14
fix issue #485 : MessagePack queries: Exception during parsing InfluxDB version [macOS]

lxhoan:


bednar:
Maybe it will be better use some implementation for the [Semantic Versioning](https://semver.org) - [semver4j](https://github.com/vdurmont/semver4j).

The method `checkMessagePackSupport` will be pretty simple:

```java
private boolean checkMessagePackSupport() {
    Semver version = new Semver(version(), Semver.SemverType.NPM);
    return version.isGreaterThan("1.4");
}
```

lxhoan:
@bednar 
I don't think main project maintainers will approve for new dependencies (semver4j) if it not indispensable

lxhoan:
there unit test added for the new version checking, it should cover MacOS case and even unrealistic cases

lxhoan:
@fmachado , please have a look on it, tks

fmachado:
@lxhoan I know this PR is already merged (and apologies for the delay) but as you requested, I'll add my comments.

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -666,13 +669,26 @@ public boolean databaseExists(final String name) {
     public String error;
   }
 
+  private boolean checkMessagePackSupport() {
+    Matcher matcher = Pattern.compile("(\\d+\\.*)+").matcher(version());
```
fmachado: AFAIK, `Pattern.compile` creates an immutable object so you can reuse it. There is another regular expression that would help you so you don't have to split: `(?:(\d+)\.)(\d+)`. (you can see it working [here](https://www.regexplanet.com/share/index.html?share=yyyyy1bypyr)). You still have to convert it to int, though.

Issue #392, opened 2017-12-04
Question: Any support for Reactive streams / RxJava integration or non-blocking alternative?


dfernandezm:
Hi all,

Sorry if this is a dumb question, I am a beginner in InfluxDB and this library. I would like to invoke it for a high volume of input coming from a Java application. That application will use RxJava / Reactive Streams all the way, so it is completely non-blocking.
Is it supported at all? Any pointers to a possible wrapping of the current calls so that they are observables or equivalent?

fmachado:
@dfernandezm IMO there is a feasible approach for reactive streams with the current API. However it's important to notice that it may not be the best solution and the API will probably be rewritten soon (for a more complete overview about existing issues, check [this document](https://github.com/influxdata/influxdb-java/wiki/Client-Improvements-Plan)).

I'm assuming that you want to write as may points in one HTTP call as possible. So I would:
1) Enable the batch mode (`org.influxdb.impl.InfluxDBImpl.enableBatch(int, int, TimeUnit, ThreadFactory, BiConsumer<Iterable<Point>, Throwable>)`);
2) Create a `BiConsumer` to notify my Observable for errors. Currently we don't have  a way to notify on successful writes (this issue is mentioned on the document).

BTW, any contribution here (e.g. a refactoring to support `CompletableFuture` or another approach that do not add external dependencies) would be great! :)

PR #493, opened 2018-08-22
Added possibility to reuse client as a core part of Reactive client

bednar:
This PR solve [#392](https://github.com/influxdata/influxdb-java/issues/392) by adding possibility to reuse influxdb-java client as a core part of [influxdb-java-reactive](https://github.com/bonitoo-io/influxdb-java-reactive) - **Reactive Java client for InfluxDB**.

#### Summary of changes
The `InfluxDBImpl` constructor has parameter for customise `Retrofit.Builder`. That builder is use in **influxdb-java-reactive** client for create reactive version of `InfluxDBService`.

### DIFF
--- a/src/main/java/org/influxdb/InfluxDBException.java
+++ b/src/main/java/org/influxdb/InfluxDBException.java
@@ -126,7 +126,7 @@ public class InfluxDBException extends RuntimeException {
     }
   }
 
-  private static InfluxDBException buildExceptionFromErrorMessage(final String errorMessage) {
+  public static InfluxDBException buildExceptionFromErrorMessage(final String errorMessage) {
     if (errorMessage.contains(DATABASE_NOT_FOUND_ERROR)) {
       return new DatabaseNotFoundException(errorMessage);
     }
--- a/src/main/java/org/influxdb/InfluxDBException.java
+++ b/src/main/java/org/influxdb/InfluxDBException.java
@@ -126,7 +126,7 @@ public class InfluxDBException extends RuntimeException {
     }
   }
 
-  public static InfluxDBException buildExceptionFromErrorMessage(final String errorMessage) {
+  private static InfluxDBException buildExceptionFromErrorMessage(final String errorMessage) {
     if (errorMessage.contains(DATABASE_NOT_FOUND_ERROR)) {
       return new DatabaseNotFoundException(errorMessage);
     }
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -107,14 +107,37 @@ public class InfluxDBImpl implements InfluxDB {
    *          The InfluxDB user name
    * @param password
    *          The InfluxDB user password
-   * @param client
+   * @param okHttpBuilder
    *          The OkHttp Client Builder
    * @param responseFormat
    *          The {@code ResponseFormat} to use for response from InfluxDB
    *          server
    */
-  public InfluxDBImpl(final String url, final String username, final String password, final OkHttpClient.Builder client,
-      final ResponseFormat responseFormat) {
+  public InfluxDBImpl(final String url, final String username, final String password,
+                      final OkHttpClient.Builder okHttpBuilder, final ResponseFormat responseFormat) {
+    this(url, username, password, okHttpBuilder, new Retrofit.Builder(), responseFormat);
+  }
+
+  /**
+   * Constructs a new {@code InfluxDBImpl}.
+   *
+   * @param url
+   *          The InfluxDB server API URL
+   * @param username
+   *          The InfluxDB user name
+   * @param password
+   *          The InfluxDB user password
+   * @param okHttpBuilder
+   *          The OkHttp Client Builder
+   * @param retrofitBuilder
+   *          The Retrofit Builder
+   * @param responseFormat
+   *          The {@code ResponseFormat} to use for response from InfluxDB
+   *          server
+   */
+  public InfluxDBImpl(final String url, final String username, final String password,
+                      final OkHttpClient.Builder okHttpBuilder, final Retrofit.Builder retrofitBuilder,
+                      final ResponseFormat responseFormat) {
     this.messagePack = ResponseFormat.MSGPACK.equals(responseFormat);
     this.hostAddress = parseHostAddress(url);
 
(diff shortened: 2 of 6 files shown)


majst01:
Hi @bednar 

You are solving 3 Issues with one PR, from my understanding only your first point is related to rxjava.

We only want to have PR´s which are focused on one single Issue.

If i am wrong can you please elaborate more details.

bednar:
Hi @majst01,

Yes, you are right. 

During implementation the reactive client I found two issues with `InfluxDBResultMapper`:
- `InfluxDBResultMapper` is not able to handle results which a different time precision then milliseconds
- `InfluxDBResultMapper` is not able to map Integer to Instant

I will split the current PR to three pull requests:

1. The `InfluxDBResultMapper` is able to map `Integer` to `Instant`
2. The `InfluxDBResultMapper` is able to handle results which a different time precision
3. Added possibility to reuse client as a core part of Reactive client

It will be ok for you?

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -146,7 +169,7 @@ public InfluxDBImpl(final String url, final String username, final String passwo
       break;
     }
 
-    this.retrofit = new Retrofit.Builder().baseUrl(url).client(
+    this.retrofit = retrofitBuilder.baseUrl(url).client(
```
fmachado: Just a heads up: you may have here `ConcurrentModificationException` as you are adding a `converterFactory` to a shared and non thread-safe `retrofitBuilder`. Worst case scenatio if you don't trigger a CME, you will end up having multiple `converterFactory` [added to the internal list in retrofitBuilder](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/Retrofit.java#L515).

src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
```diff
@@ -152,7 +202,7 @@ void throwExceptionIfResultWithError(final QueryResult queryResult) {
     });
   }
 
-  void cacheMeasurementClass(final Class<?>... classVarAgrs) {
+  public void cacheMeasurementClass(final Class<?>... classVarAgrs) {
```
fmachado: Please revert the access modifier from this method on this class and from all methods from other classes to the minimum required. There is no need to make this method public.

src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
```diff
@@ -172,13 +222,22 @@ void cacheMeasurementClass(final Class<?>... classVarAgrs) {
     }
   }
 
-  String getMeasurementName(final Class<?> clazz) {
+  public String getMeasurementName(final Class<?> clazz) {
     return ((Measurement) clazz.getAnnotation(Measurement.class)).name();
   }
 
+  public <T> ConcurrentMap<String, Field> getColNameAndFieldMap(final Class<T> clazz) {
```
fmachado: There is no need to have this method: you are not mocking it and now you are exposing an internal structure publicly to the external code, allowing it to be changed. **Please, revert this change.**

src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
```diff
@@ -263,9 +324,11 @@ String getMeasurementName(final Class<?> clazz) {
       if (value instanceof String) {
         instant = Instant.from(ISO8601_FORMATTER.parse(String.valueOf(value)));
       } else if (value instanceof Long) {
-        instant = Instant.ofEpochMilli((Long) value);
+        instant = Instant.ofEpochMilli(toMillis((Long) value, precision));
```
fmachado: Could you please do a tinny improvement here by casting to primitives instead of Object wrappers? If you do, then please change the method signature from `toMillis(final Long value, ...` to `toMillis(final long value, ...` (just to avoid unboxing).

src/main/java/org/influxdb/impl/InfluxDBImpl.java
bednar: Yes, you are right. It could cause same errors as in #445, so I will use same approach to avoid potential problems:

```java
Retrofit.Builder clonedRetrofitBuilder = retrofitBuilder.baseUrl(url).build().newBuilder();
this.retrofit = clonedRetrofitBuilder.client(clonedOkHttpBuilder.build())
            .addConverterFactory(converterFactory).build();
```

bednar:
@fmachado I made the required changes. Can you check it?

majst01:
Hi @bednar 

This PR still modifies around precision handling, and instantiation of a new influxdb service. I doubt these are all related to the usage in a reactive environment.

As long as we have mixed modifications in this PR we will not accept that.

bednar:
Hi @majst01, thanks for response. 

Today, I will create separate PR for handling precision and leave this PR only for instantiation of a new InfluxDB service.

bednar:
Hi @majst01, @fmachado 

I created pull request (#501) for the InfluxDBResultMapper precision issue and also changed description of current PR to reflect changes.

Thanks for your reviews

src/main/java/org/influxdb/InfluxDBException.java
```diff
@@ -126,7 +126,7 @@ public boolean isRetryWorth() {
     }
   }
 
-  private static InfluxDBException buildExceptionFromErrorMessage(final String errorMessage) {
+  public static InfluxDBException buildExceptionFromErrorMessage(final String errorMessage) {
```
majst01: Why public, if you want to test this like below, write the test in the matching package.

majst01: Please add a changelog.md entry with pointers where the usage of influxdb-java in a reactive environment is shown.

bednar: Because I want to reuse this method in influxdb-java-reactive to parse errors from HTTP header X-Influx-Error. If its is a problem I can use something like this:

```java
String errorBody = String.format("{\"error\":\"%s\"}", errorHeader);
```

Can I let this method public?

bednar: The CHANGELOG.md was updated.

bednar:
Hi @majst01,

I changed method InfluxDBException#buildExceptionFromErrorMessage to private.

Thanks for your review

PR #496, opened 2018-08-27

lxhoan:
This PR is to resolve host address for UDP writing, period is 5 mins
(follow the default eviction period of OKHttp conection pool)


lxhoan:
This PR is to resolve host address for UDP writing, period is 5 mins
(follow the default eviction period of OKHttp conection pool)

PR #502, opened 2018-08-29

lxhoan:
this PR is a simpler solution for #480

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -4,7 +4,6 @@ package org.influxdb.impl;
 import com.squareup.moshi.JsonAdapter;
 import com.squareup.moshi.Moshi;
 import okhttp3.Headers;
-import okhttp3.HttpUrl;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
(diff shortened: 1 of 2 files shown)


lxhoan:
Hi @fmachado , @majst01 , in case we have dns resolving cached at JVM level, still need this PR to fix issue #480

lxhoan:
Hi @majst01 , @fmachado
What do you think ?
tks

majst01:
Hi @lxhoan 

I like this approach, but we should at least measure what performance impact this gives ?

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -187,18 +188,20 @@ public InfluxDBImpl(final String url, final String username, final String passwo
     setRetentionPolicy(retentionPolicy);
   }
 
-  private InetAddress parseHostAddress(final String url) {
-      HttpUrl httpUrl = HttpUrl.parse(url);
+  private String parseHost(final String url) {
+    HttpUrl httpUrl = HttpUrl.parse(url);
```
fmachado: Is `HttpUrl` able to parse URIs that do not start with the string `http`? IMO we should treat this parameter as an `URI` and not an `URL`. The main reasons are:
* Users may finally stop using the scheme "http://" to connect via UDP. [RFC 1738 for URL](https://tools.ietf.org/html/rfc1738#section-3) does not allow the scheme "udp://" but the [RFC 3986 for URI](https://tools.ietf.org/html/rfc3986#section-3.1) does.
* If so, this line could be replaced with `URI.create(...)` and later you could use `URI#getHost()` to get only the hostname.

lxhoan:
@fmachado , I have fixed as your comment, tks

lxhoan:
Hi @majst01 , 

>I like this approach, but we should at least measure what performance impact this gives ?

Performance impact should be inconsiderable since DNS resolving is cached at JVM level (as we discussed intensively before), so most of ```InetSocketAddress``` resolving is are just intra-process invocations

lxhoan:
closed (PR #499 as an alternative)

Issue #439, opened 2018-04-02
Retry capability for writing of BatchPoints as well


lxhoan:
InfluxDB v 2.9 shipped with retry capability feature for write(final Point point)
I believe we should have the same feature for write(final BatchPoints batchPoints)

PR #503, opened 2018-08-29
imlement issue #439 : Retry capability for writing of BatchPoints

lxhoan:
I  believe in the value of support BatchPoints writing with retry
This PR is to expose one method for this purpose


lxhoan:
this PR as an alternative to PR #495 
This looks simpler because the purpose is exposed directly in InfluxDB interface

Please take a look, tks

lxhoan:
Hi @majst01 , @fmachado 
What do you think ?

Issue #505, opened 2018-08-29
Error messages from server not parsed correctly when using msgpack


mschwore:
I am using the new msgpack format with influxdb-java and I am not seeing the right exceptions being thrown by influxdb-java. Here's a stack trace for an exception caught by my app:

```
org.influxdb.InfluxDBException: ??error?9partial write: points beyond retention policy dropped=463
        at org.influxdb.InfluxDBException.buildExceptionForErrorState(InfluxDBException.java:168) ~[influxdb-java-2.12.jar:?]
        at org.influxdb.impl.InfluxDBImpl.execute(InfluxDBImpl.java:690) ~[influxdb-java-2.12.jar:?]
        at org.influxdb.impl.InfluxDBImpl.write(InfluxDBImpl.java:422) ~[influxdb-java-2.12.jar:?]
        at org.influxdb.impl.OneShotBatchWriter.write(OneShotBatchWriter.java:22) ~[influxdb-java-2.12.jar:?]
```
I expected to see an exception of type PointsBeyondRetentionPolicyException. It looks like the InfluxDBException class is attempting to parse the error message from influxdb-server as JSON and failing as influxdb-server is sending back a msgpack formatted error message.

Here I am just using the OneShotBatchWriter, but I imagine if you used the retry batch writer that the retry capability will be broken.

mschwore:
This was using influxdb-java 2.12, influxdb-server 1.6.2 on CentOS 7

majst01:
@mschwore can we close this issue ?

PR #506, opened 2018-08-30
fix issue #505 : Error messages from server not parsed correctly when using msgpack

lxhoan:


lxhoan:
Hi @majst01 , @fmachado 
all pre-checks have passed, please help to review, tks

Issue #498, opened 2018-08-27
InfluxDBException: 'error parsing query' - Error is not propagated to Consumer callback


anastasiaO:
Hello,

please take a look at [InfluxDBImpl.java#L548](https://github.com/influxdata/influxdb-java/blob/master/src/main/java/org/influxdb/impl/InfluxDBImpl.java)

If you throw an InfluxDBException an this position, it is not possible to catch it on the caller side:

```
public void loadDataAsync(int chunk, Consumer<QueryResult> consumer) {
    // Bad syntax test
    db.query(new Query("SELECT * FROM sieries LIMIT -100", databaseName), chunk, consumer);
}
```

Stacktrace:

> org.influxdb.InfluxDBException: {"error":"error parsing query: found -, expected integer at line 1, char 63"}
> 
> 	at org.influxdb.impl.InfluxDBImpl$1.onResponse(InfluxDBImpl.java:420)
> 	at retrofit2.OkHttpCall$1.callSuccess(OkHttpCall.java:138)
> 	at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:117)
> 	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:141)
> 	at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
> 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
> 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
> 	at java.lang.Thread.run(Thread.java:748)

This causes my REST-Endpoint (Spring MVC) to hang.
![image](https://user-images.githubusercontent.com/10603718/44657325-bba7f700-a9fc-11e8-8bec-1d496b6ce027.png)

PR #509, opened 2018-08-31
fix issues 498: Added streaming with time units to the influx db interface

weiss19ja:
Fix for issues #498 to propagate exception if you are using the query stream.

Adds also functionality to specify the time unit to the query stream. This feature was also disred by #418.
It is needed if you want to stream a great among of data and don't want to parse the dates from the ISO format to an epoch timestamp.


majst01:
Hi @weiss19ja long time no update, if still interested, please rebase.

PR #510, opened 2018-08-31
Feature/query builder

gkatzioura:
This is an implementation of an advanced query builder for influxdb.
The user is able to create queries by specifying the clauses like equals, non equals, containing, aggregations etc. Pretty much in the same way when he uses jooq.

Full use cases are displayed on the BuiltQueryTest.java class.

I was implementing the influxdb spring data repository and thus I needed some extra functionality.
The query builder was inspired from the cassandra java driver.
Packaging it with the influxdb driver made more sense to me than packaging it in another binary.

All checks are passed. 
There is a readme on the usage and tests on every use case.
@majst01 , @fmachado
What do you think?

Some examples

```java
    Query select =
        select()
            .all()
            .from(DATABASE, "foobar")
            .where(eq("test1", 4))
            .and(gt("test2", "a"))
            .and(lte("test2", "z"))
            .orderBy(asc());
```

```java
    Query select =
        select()
            .max("test1")
            .min("test2")
            .count("test3")
            .sum("test4")
            .from(DATABASE, "foobar")
            .limit(1, 20);
```

```java
Query select = select().function("MEDIAN", "test1").from(DATABASE, "foobar").limit(1, 20);
```

```java
    Query select =
        select()
            .column("test1")
            .from(DATABASE, "foobar")
            .where(eq("test1", 1))
            .or(eq("test2", "a"))
            .or(eq("test3", "b"))
            .and(eq("test4", "c"));
```

```java
Query select =
                select()
                        .column("test1")
                        .from(DATABASE, "foobar")
                        .where(eq("test1", 1))
                        .or(eq("test2", "a"))
                        .or(eq("test3", "b"))
                        .andNested()
                          .and(eq("test2","b"))
                          .or(eq("test3","a"))
                        .close()
                        .orNested()
                          .and(eq("test1",2))
                          .and(eq("test2","y"))
                          .and(eq("test3","z"))
                        .close()
                        .andNested()
                          .or(eq("test1",8))
                          .or(eq("test2","g"))
                          .or(eq("test3","j"))
                        .close()
                        .and(eq("test4", "c"));
```


majst01:
Hi @gkatzioura 

Great, appreciate your effort, amazing ! I was thinking about something like that to add at the beginning, but hadnt time to implement ever. So this would mark a big step forward.

I would like to add this in the next major version (2.14), adding it now is a bit to late as we are in the last steps of releasing 2.13.

@fmachado what do you think ?

gkatzioura:
Adding it on the next major version would be great!
Once the time for the release comes, it might come with some extra changes and enhancements, since it is already used, thus more feedback.
Until released is it going to be in a snapshot or just the source code?
Thank you!

lxhoan:
With reference to
https://docs.influxdata.com/influxdb/v1.6/query_language/data_exploration/

The approach looks missing:
+ Subqueries
+ Time arithmetic
+ Time zone clause
+ SOffset and SLimit clause
+ INTO clause

In general the missing (if I am correct) should be documented otherwise it can make user confused when to use it

gkatzioura:
Thanks for the feedback @lxhoan!
I will implement and try to minimize the non implemented ones from the list.
The ones missing will be documented.

Just to mention that in case a feature is not implemented, which is highly likely if a new influxdb query feature rolls out, the user is still able to use the query builder, by using the raw text functionality
```java
 select() .column("test1")
    .column("test2")
    .all()
    .from(DATABASE, "foobar")
    .where(ne("test1", raw("raw expression"))),
```

```java
select()
    .raw("an expression on select")
    .from(DATABASE, "foobar")
    .limit(1, 20);
```

gkatzioura:
Hi!

Just wanted to catch up, and give some status on what has been done. It seems this weekend it will be finalized in terms of features.

I added support for subqueries

```java
    Query select =
        select()
            .requiresPost()
            .column("column1")
            .column("column2")
            .fromSubQuery(DATABASE)
              .max("column1")
              .max("column2")
              .fromSubQuery()
                .all()
                .fromSubQuery()
                  .max("column1")
                  .mean("column2")
                  .fromSubQuery()
                    .column("test1")
                    .distinct()
                    .from("foobar")
                   .where(gt("column1", 3))
                    .groupBy("column2")
                    .limit(1, 20)
                    .sLimit(2, 10)
                    .close()
                .close()
                .where(eq("column1", 5))
                .groupBy("column2")
                .limit(50, 10)
                .close()
              .where(eq("column1", 4))
              .sLimit(3)
              .close()
            .where(eq("column3", 5))
            .limit(5, 10);
```

Support for slimit and soffset

```java
Query select =
        select().column("test1").from(DATABASE, "foobar").groupBy("test2", "test3").sLimit(1, 20);

```

Support for the time zone clause

```java
    Query select =
        select()
            .column("test1")
            .from(DATABASE, "foobar")
            .groupBy("test2", "test3")
            .sLimit(1)
            .tz("America/Chicago");
```
Time arithmetic on group by.

```java
    Query select = select().column("test1").from(DATABASE, "foobar").groupBy(time(1l, HOUR, 5l, WEEK));
```

Soon I will also have the INTO clause implemented and thus a Readme updated with the query builder capabilities.

gkatzioura:
Greeting!
I’ve just finished with the new version of the query builder.
To make sure that all the features are supported I went through 
The https://docs.influxdata.com/influxdb/v1.6/query_language/data_exploration
and wrote examples for each use case specified by using the query builder.

The readme section on the query builder contains these examples. 

https://github.com/influxdata/influxdb-java/blob/90e42ecf1a22909dc23583b2c0dab58839a91daa/README.md

Thus the user has enough examples on how to use the query builder for almost every select use case. Another concern is the performance therefore the query builder does one concatenation per query statement.

@majst01  , @fmachado , @lxhoan 
What do you think?

Kind Regards!

gkatzioura:
Hi! 
I am sending this message to catch up and mention that I am open for a call/discussion and get some feedback or adjustments in order to make the query builder feature bring as much value possible to the influxdb’s java driver users. 

King regards,
Emmanouil

majst01:
Hi @gkatzioura 
one last nit before merging, can you put the README.md parts into a seperate document and link it in the README.md. I'm afraid of such a long file, it would prevent a lot of people from reading.

Thanks

Issue #513, opened 2018-09-06
Response body must be closed properly in case of JSON response


lxhoan:
async Query with chunking does not handle RuntimeException and this case the response body will not closed

PR #514, opened 2018-09-06
fix issue #513 : Response body must be closed properly in case of JSON response

lxhoan:


lxhoan:
hi @majst01 , @fmachado , this seems a valid defect, please help to take a look, tks

PR #519, opened 2018-09-11
Features/add fields by pojo

mirza99:
Enable adding fields by passing a POJO to the Builder.

### How to use it

```
// Order is a POJO having annotations @Measurement and @Column
Order order = new Order();
// ... setting data

Point p = Point.measurementByPOJO(order.getClass()).addFieldsFromPOJO(order).build();
```


src/main/java/org/influxdb/dto/Point.java
```diff
@@ -196,6 +225,104 @@ public boolean hasFields() {
...
+
+      return this;
+    }
+
+    private boolean isPublic(final Field field) {
+      int modifiers = field.getModifiers();
+      return Modifier.isPublic(modifiers);
+    }
+
+    private void addFieldByMethod(final Object pojo, final Class<? extends Object> clazz,
```
fmachado: If I'm not wrong, this implements a different behavior from the initial toPOJO implementation where we get data directly from the field instead of unnecessarily calling the accessor method. Is there any good reason to have this call?

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -196,6 +225,104 @@ public boolean hasFields() {
...
+
+        Column column = field.getAnnotation(Column.class);
+
+        if (column == null) {
+          continue;
+        }
+
+        String fieldName = column.name();
+
+        if (isPublic(field)) {
```
fmachado: I don't think it matters if the field is public or not: you should get the field value anyway as it was annotated with (at)Column annotation.

mirza99: I've tested it, if it's private, it cannot be accessed. 
I think I can set `accessible` to true. I'll test it and update the pull request.

fmachado:
@mirza99 I would like to merge this PR but some checks are failing. Would be possible to fix them? Thanks again for your contribution!

mirza99:
@fmachado You can merge the PR now. Sorry for the delay.

Issue #521, opened 2018-09-17
Memory leak in StringBuilder cache for Point.lineprotocol()


rhajek:
Hello, 

Point.CACHED_STRINGBUILDERS is per thread cache to store a StringBuilder instances for constructing lineprotocol.  As the key in the map is used measurement name, the value is StringBuilder instance. The instance of StringBuilder is reused for all points with the same measurememnt. 

This optimalization reduces GC activity when writing large number of Points but increases the memory consumption when we have large number of different measurements. The cache can memory leake, because it is never evicted programmatically and it is garbage collected only when the holding thread dies. StringBuilder.setLength(length) is used for reusing instance, but is does not free the builder memory.

OutOfMemory can be reproduced by writing large number of Points with unique measurements with long line protocol length in one thread.

majst01:
Yes, but do you have a idea to prevent that situation ? Having a massive amount of measurements is of course memory intensive, but the performance gain we have actually is massive.

rhajek:
I tried to replace 
```java 
private static final ThreadLocal<Map<String, MeasurementStringBuilder>>
```
with simple thread local shared StringBuilder and it performs better cca 10% for large number of unique measurements (20k+). Performance for less complex lineprotocol is not degraded. I used inch-java to measure this. 

```java
private static final ThreadLocal<StringBuilder> CACHED_STRINGBUILDERS
```
Version without thread local map: 
```
Host: http://localhost:8086
Client version: 2.14-sb2-SNAPSHOT
InfluxDB server version: 1.6.2
Concurrency: 5
Measurements: 50000
Tag cardinalities: [2, 2, 2]
points per series: 2
Total series: 400000
Total points: 800000
Total fields per point: 3
Batch Size: 5000
Database: j_stress (Shard duration: 7d)
Write Consistency: any
BatchPoints set prepared.
Start time: 1537185957047000000
Approx End time: 1537189557047000000
T=0 0 points written. Total throughput: � pt/sec | � val/sec. Current throughput: 0 val/sec Memory usage: 1316784KB
T=1 10,000 points written. Total throughput: 10,000 pt/sec | 30,000 val/sec. Current throughput: 10,000 val/sec Memory usage: 1309640KB
T=3 50,000 points written. Total throughput: 16,666.666 pt/sec | 50,000 val/sec. Current throughput: 40,000 val/sec Memory usage: 1248597KB
[...]

PR #522, opened 2018-09-17
- fixing memory leak and performance #521

rhajek:
Fixing https://github.com/influxdata/influxdb-java/issues/521
ThreadLocal  StringBuilder hashmap cache was replaced by simple ThreadLocal StringBuilder.

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -320,25 +320,25 @@ public class Point {
     return lineProtocol(null);
   }
 
-    /**
-     * Calculate the lineprotocol entry for a single point, using a specific {@link TimeUnit} for the timestamp.
-     * @param precision the time precision unit for this point
-     * @return the String without newLine
-     */
-    public String lineProtocol(final TimeUnit precision) {
-      final StringBuilder sb = CACHED_STRINGBUILDERS.get();
-      sb.setLength(0);
+  /**
+   * Calculate the lineprotocol entry for a single point, using a specific {@link TimeUnit} for the timestamp.
+   * @param precision the time precision unit for this point
+   * @return the String without newLine
+   */
+  public String lineProtocol(final TimeUnit precision) {
 
-      escapeKey(sb, measurement);
-      concatenatedTags(sb);
-      concatenatedFields(sb);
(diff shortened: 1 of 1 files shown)


src/main/java/org/influxdb/dto/Point.java
```diff
@@ -336,15 +326,18 @@ public String lineProtocol() {
...
-              .resetForUse();
+      final StringBuilder sb = CACHED_STRINGBUILDERS.get();
+      sb.setLength(0);
 
-        concatenatedTags(sb);
-        concatenatedFields(sb);
+      escapeKey(sb, measurement);
+      concatenatedTags(sb);
+      concatenatedFields(sb);
+      if (precision != null) {
```
fmachado: This check is not necessary. `formatedTime` is already doing it.

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -336,15 +326,18 @@ public String lineProtocol() {
      * @return the String without newLine
      */
     public String lineProtocol(final TimeUnit precision) {
-      final StringBuilder sb = CACHED_STRINGBUILDERS
-              .get()
-              .computeIfAbsent(this.measurement, MeasurementStringBuilder::new)
-              .resetForUse();
+      final StringBuilder sb = CACHED_STRINGBUILDERS.get();
+      sb.setLength(0);
```
fmachado: Just to be kind to the next developer who wonder why we do this here, I would add a comment explaining our intention.

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -336,15 +326,18 @@ public String lineProtocol() {
      * @return the String without newLine
      */
     public String lineProtocol(final TimeUnit precision) {
-      final StringBuilder sb = CACHED_STRINGBUILDERS
-              .get()
-              .computeIfAbsent(this.measurement, MeasurementStringBuilder::new)
-              .resetForUse();
+      final StringBuilder sb = CACHED_STRINGBUILDERS.get();
```
fmachado: IMO declaring variables as final if they are (1) only visible inside a method and (2) not strictly required to be final only bloats the code.

Issue #528, opened 2018-10-02
Fails to parse host address when running in Docker swarm environments


lindsve:
We upgraded from 2.10 to 2.13 and noticed that we could not get our application running in our docker swarm environment, because it fails when connecting to Influx:

`java.lang.IllegalArgumentException: Unable to parse url: http://swarm_influx:8086`

This works in 2.11 which uses HttpUrl to parse host names, but I see that this is changed recently to use `java.net.Uri` instead.

Docker swarm autogenerates host urls based on the scheme `<name-of-swarm-stack>_<name-of-service>` so this is out of our control unfortunately.

We have currently downgraded to 2.11

majst01:
According to: https://en.wikipedia.org/wiki/Hostname and https://www.ietf.org/rfc/rfc1912.txt underscores (_) are not allowed in hostnames.

lindsve:
I know, it's basically a PITA that docker uses this url scheme. There's a registered issue for it, so hopefully they will be more compliant in the future https://github.com/moby/moby/issues/37632

On that note, I've seen underscores in hostnames for cloud provided solutions as well, so it's seemingly not uncommon although it's not per spec.

majst01:
Yes but with this bug in the moby runtime, all Java based applications will fail, so they have to fix this, it cannot be fixed in influxdb-java.

fmachado:
I found a workaround that may help for you: https://github.com/sitewhere/sitewhere/issues/539#issuecomment-328654010
If this works, could you please add your comments here? (I will keep this issue closed because this is **not** an influxdb-java library).

As far as I could understand, this is an old issue (2014 - github.com/docker/compose/issues/229) caused by how Docker (compose/swarm) creates a hostname.

PR #529, opened 2018-10-02
- fixed issue #528, hostname can contain "_" character

rhajek:
Replacing `java.net.URI` with  newer `java.net.URL` fixed this.

https://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn/176274#176274

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -42,9 +42,9 @@ import java.net.DatagramPacket;
 import java.net.DatagramSocket;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.net.MalformedURLException;
 import java.net.SocketException;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.net.URL;
 import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


rhajek:
There is a problem with code coverage: `new URL(url).getHost()` never returns `null`. It is not possible to reproduce it in test. I am not sure that it behaves the same on all implementations JVM, android...

if (hostName == null || "".equals(hostName)) {
      throw new IllegalArgumentException("Unable to parse url: " + url);
}

I would like to let that check here for sure.. I think we can accept small codecov decrease  in this case.

fmachado:
IMO this PR is (1) not fixing but reverting to the previous wrong implementation and (2) accepting "_" as part of the hostname when @majst01 said, I explained on the initial implementation by @lxhoan, Oracle said in https://bugs.java.com/bugdatabase/view_bug.do?bug_id=5049974 and it is stated in multiple RFCs that is not allowed.

I don't know how to make this more clear.

fmachado:
AFAIK "udp://127.0.0.1" is not an URL but an URI (accordingly with RFCs).
If I'm connecting to InfluxDB via UDP. should I use "http://127.0.0.1"? Because this is what your implementation will return to.

rhajek:
It is not ideal solution, it is workaround for real world use case (docker). Unfortunately using (_) in domain name is a common issue. Strictly in RFC is using "_" is forbidden in hostname, but it is legal in subdomain name.
https://stackoverflow.com/questions/2180465/can-domain-name-subdomains-have-an-underscore-in-it

"udp://127.0.0.1" is also not complete URI - there is unknown port, udp not the protocol - it is more like transport. Similarly you can't connect to "tcp://127.0.0.1".

For me the best solution is to parse "udp://" using URI and "http[s]" by URL.

majst01:
Hi @rhajek we are talking about hostnames with underscores, and they are not allowed, i am not willing to merge wrong implementations into a library because there are bugs in other software solutions out there, i will close this PR

PR #530, opened 2018-10-03
simplify and code style

heshengbang:


majst01:
Hi @heshengbang.
what exact problem are you trying to solve ? changing ~900 lines of code and broken tests, i will close this.

PR #531, opened 2018-10-05
Easy to use try-with-resources

heshengbang:
Easy to use try-with-resources


heshengbang:
Hi @majst01 ,
  I have added README.md and move AutoCloseable to InfluxDB that's the place it should be more suitable.

PR #533, opened 2018-10-12
Add save and load functionalities using measurement models

gkatzioura:
Hi! I needed a save and load object only functionality thus I created some utils in order to persist and load measurement models directly.

Just like the [DynamoDBMapper](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.html) in java I believe it would be a great feature to add to influxdb.

To the measurement annotation I added information such as the retention policy and the database

I made a class that inherits the InfluxDBResultMapper. I think it can be the same in the first place.

I added the query measurement by the model and saving them by model 

What might be confusing is saving by providing the time. It can be instant only but this has to do with the result mapper parsing it to Instant. If no time column is provided it will use the system time.

Some examples follow

```java
@Measurement(name = "server_measure", database = UDP_DATABASE, retentionPolicy = "default")
static class ServerMeasure {

  /** Check the instant convertions */
  @Column(name = "time")
  private Instant time;

  @Column(name = "name", tag = true)
  private String name;

  @Column(name = "cpu")
  private double cpu;

  @Column(name = "healthy")
  private boolean healthy;

  @Column(name = "min")
  private long uptime;

  /** TODO bigdecimal unsupported? */
  @Column(name = "memory_utilization")
  private Double memoryUtilization;
```

Save it

```java
ServerMeasure serverMeasure = new ServerMeasure();
serverMeasure.setName("maverick");
serverMeasure.setCpu(4.3d);
serverMeasure.setHealthy(true);
serverMeasure.setUptime(1234l);
serverMeasure.setMemoryUtilization(new Double(34.5));
influxDBMapper.save(serverMeasure);

```

Load using the model

```java
List<ServerMeasure> persistedMeasure = influxDBMapper.query(ServerMeasure.class);
```

Load using the query

```java
Query query = ..create your query
List<ServerMeasure> persistedMeasure = influxDBMapper.query(query,ServerMeasure.class);
```

Kind Regards

### DIFF
--- a/INFLUXDB_MAPPER.md
+++ b/INFLUXDB_MAPPER.md
@@ -1,6 +1,6 @@
 ### InfluxDBMapper
 
-In case you want to use models only you can use the InfluxDBMapper to save and load measurements.
+In case you want to use models only, you can use the InfluxDBMapper to save and load measurements.
 You can create models that specify the database the measurement and the retention policy.
 
 ```Java
@@ -25,7 +25,7 @@ public class Cpu {
 Save operation using a model.
 
 ```Java
-Cpu cpu = ..create the cpu measure
+Cpu cpu = .., create the cpu measure
 influxDBMapper.save(cpu);
 ```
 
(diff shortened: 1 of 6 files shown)


majst01:
Hi @gkatzioura 

great work, like it, but same as for the querybuilder i would opt for a seperate documentation with link from README.md

PR #552, opened 2019-01-21
Database name is optional for Query and BatchPoints.

### LINKED ISSUE #548: Database name when using InfluxDB
Every time I want to run a Query or write some batch points I need to pass the database name, although it is already configured in the InfluxDB instance. E.g:
0.  `InfluxDB influxDb = InfluxDBFactory.connect(url, username, password).setDatabase(DB_NAME_INFLUX);`
1. `BatchPoints batchPoints = BatchPoints.database(DB_NAME_INFLUX)` 
2. `new Query(QUERY_STRING, DB_NAME_INFLUX)`

and then 
`influxDb .query(query); //but influxDB is already configured with a DB name` 

Basically one needs to pass it over and over again and it is a design smell to me. 
A solution could be that queries and BatchPoints do not require a database name, and this is set on execution by the InfluxDB instance if empty. Other database APIs where a Connection-instance (i.e `InfluxDB` instance) per database is forced seem more clear.

asashour:
Fixes #548

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## 2.15 [unreleased]
 
+### Features
 
+- Query and BatchPoints do not mandate a database name, in which case the InfluxDB database
+  would be used [Issue #548](https://github.com/influxdata/influxdb-java/issues/548)
 
 ## 2.14 [2018-10-12]
 


majst01:
With this end user api breaks right ? Any Ideas howto prevent that

asashour:
No, user API doesn't break, there are no methods removed. We just add new methods.  (I guess old methods should be deprecated).

PR #557, opened 2019-01-21
Better spacing.

asashour:


majst01:
Hi,
i considered doing stuff like this, but i would prefer to have the rules defined in the checkstyle then forces to break the build, same applies to #558 .

If you are aware of code formatting tools which work and and not to intrusive this would be appreciated.

i am not going to pull this and #558 

But thanks for all your other effort, appreciate !

asashour:
I understand, but just a hint, adding the tests sources currently gives 898 errors, and reviewing them in one PR would be huge.

I am creating a new PR, with steps of commits per type of checkstyle, please stop me when it is enough.

PR #560, opened 2019-01-22
Checkstyle

asashour:
Include test sources, and fix each checkstyle type.


asashour:
Should I close this PR? Appreciate your feedback.

asashour:
In other words, after this PR, any violation of checkstyle fails the build, and mandates the code in the test package conforms to the specified checkstyle.

majst01:
Yeah, but the formatting is not automated

asashour:
I find it a little stretch, since this means the build server needs to recommit the formatted code.

Anyway, I would propose a PR to include that plugin, and we check its conformance to fail the build.

I hope the auto format doesn't affect code readability.

majst01:
No not recommit, just break if it would need to format, but a maven target which can be used by every developer to format the code by our convention, which does not depend on the developers IDE

PR #561, opened 2019-01-22
Add BatchPoints.Builder.points(Collection)

### LINKED ISSUE #451: Request for new method accepting a Collection<Point> parameter in BatchPoints.Builder class
Currently I see that the BatchPoints.Builder class has a method (with parameter Point ...) to add multiple points to the BatchPoints. It is converting the variable argument parameter into a list so that all the points can be added.

I request for a new method that directly accepts a Collection<Point> as parameter which would avoid a conversion to array and back in a scenario of having a proper Collection object.

asashour:
Fixes #451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Query and BatchPoints do not mandate a database name, in which case the InfluxDB database
   would be used [Issue #548](https://github.com/influxdata/influxdb-java/issues/548)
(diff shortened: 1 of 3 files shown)


majst01:
Thanks, also worth a CHANGELOG.md entry, if added i will pull

asashour:
Thanks, done.

As a side node, I guess many issues should be closed, as they have been replied, or stated as invalid.

PR #562, opened 2019-01-22
@Column to support support super class inheritance

### LINKED ISSUE #367: @Column does not support super class inheritance
When I use the above annotation on a super class on an field that is to be inherited by another class, the annotation is not inherited to the child class as expected. The @Inherited meta-annotation needs to be added to it.

asashour:
Fixes #367


majst01:
Can you add a Changelog.md entry please, i will pull then.

PR #569, opened 2019-01-25
ResponseBody should be checked

WangSZ:
ResponseBody should be checked for error maybe


majst01:
can you describe which problem you are facing ? Other than that, this PR reduces coverage, i`d like to have always to enlarge coverage for new code. So the introduced code path must be covered.

majst01:
lot of code added, but no test for them, i close this for now, please reopen any time if you wish

Issue #573, opened 2019-02-08
InfluxDBResultMapper.toPOJO(...) after query with tz() clause throws java.time.format.DateTimeParseException


gachdn:
### Problem:
When adding the tz() clause to a working query the InfluxDBResultMapper.toPOJO(...) cannot parse the time String returned by Influx. Please see details below.

### Versions: 
- InfluxDb 1.7.1
- influxdb-java 2.14 

### Possible solution:
line 62 in _org.influxdb.impl.InfluxDBResultMapper.java_ defines the format  with the .appendPattern("X").
The **"X"** pattern only parses 'Z' or '-18' to '+18'

putting **"XXX"** pattern solved the parsing issue -**BUT**- i have no ideas which other formats are returned by InfluxDB...

DateTimeFormatter Javadoc:
> Offset X and x: This formats the offset based on the number of pattern letters. One letter outputs just the hour, such as '+01', unless the minute is non-zero in which case the minute is also output, such as '+0130'. Two letters outputs the hour and minute, without a colon, such as '+0130'. Three letters outputs the hour and minute, with a colon, such as '+01:30'. Four letters outputs the hour and minute and optional second, without a colon, such as '+013015'. Five letters outputs the hour and minute and optional second, with a colon, such as '+01:30:15'. Six or more letters throws IllegalArgumentException. Pattern letter 'X' (upper case) will output 'Z' when the offset to be output would be zero, whereas pattern letter 'x' (lower case) will output '+00', '+0000', or '+00:00'. 

### Traces:
- JSON Response without **tz('Europe/Berlin')** clause: ( toPOJO(...)  works fine ! )
`...
			"columns": ["time",
			"minmax",
			"firstlast"],
			"values": [["2019-01-01T00:00:00Z",
			37,
			37],
			["2019-01-02T00:00:00Z",
			202,
			202],
...
`

- JSON Response with **tz('Europe/Berlin')** clause: ( throws exception )
`...
			"columns": ["time",
			"minmax",
			"firstlast"],
			"values": [["2019-01-01T00:00:00+01:00",
			37,
			37],
			["2019-01-02T00:00:00+01:00",
			202,
			202],
...
`

- The Exception:
`java.time.format.DateTimeParseException: Text '2019-01-01T00:00:00+01:00' could not be parsed, unparsed text found at index 22
	at java.time.format.DateTimeFormatter.parseResolved0(Unknown Source)
	at java.time.format.DateTimeFormatter.parse(Unknown Source)
[...]

PR #575, opened 2019-02-19
Fix error parsing timestamp containing offset (#573)

fmachado:
The new pattern will parse timestamp containing zone ids or offsets.
It's also compatible with the subtle difference between ISO8601 and
RFC3339 (offsets with +00:00 and -00:00).


majst01:
Hi @fmachado can you merge my last commit from master, this fixes this annoying jdk11 javadoc bug, this is required that the build succeeds.

PR #587, opened 2019-04-08
bug-> org.influxdb.InfluxDBException$UnableToParseException: invalid …

kuhnwei:
…number;

      Because "field.get(java.time.Instant)" is not long type

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -5,6 +5,7 @@ import java.lang.reflect.Field;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.text.NumberFormat;
+import java.time.Instant;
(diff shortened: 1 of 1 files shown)


src/main/java/org/influxdb/dto/Point.java
```diff
@@ -262,6 +263,9 @@ private void addFieldByAttribute(final Object pojo, final Field field, final Col
         if (column.tag()) {
           this.tags.put(fieldName, (String) fieldValue);
         } else {
+          if ("time".equals(fieldName) && fieldValue instanceof Instant) {
```
majst01: Hi, i dont want to have hard coded field names in the Point object, you need to find a other way to ensure this is de/serialized in the correct way.

majst01:
No response, i will close this PR, feel free to reopen any time,

Issue #581, opened 2019-03-20
What is the correct way ro create a database ?


thebuleon29:
Hi,
All the examples use the method influxDB.createDatabase, but it is deprecated. What is the new method that replaces createDatabase ?

Issue #524, opened 2018-09-25
Method createDatabase is deprecated


gentixharavina:
So i need to use the method createDatabase() which is deprecated but i cannot find another alternative solution to creating a database, any suggestions?

Help please!

alisanguo:
I have the same question, any solutions?

ysfseu:
the same issue? any alternative?

gkatzioura:
You need to create the database using a query
```
this.influxDB.query(new Query("CREATE DATABASE databasename","databasename"));
```

fmachado:
```Java
  /**
   * Create a new Database.
   *
   * @param name
   *            the name of the new database.
   * @deprecated (since 2.9, removed in 3.0) Use <tt>org.influxdb.InfluxDB.query(Query)</tt> to execute a parameterized
   *             <strong>CREATE DATABASE</strong> query.
   */
  @Deprecated
  public void createDatabase(final String name);
```

PR #591, opened 2019-04-22
Update README.md

sfeilmeier:
- Replace deprecated createDatabase()-call. #524 #581
- Add missing bracket


majst01:
we should check our unit-test for the same

PR #597, opened 2019-05-24
pojo to timestamp field

flashmouse:
i add a new annotation for timestamp so that we can set point time and precision attribute when use addFieldsFromPOJO method,and also can reslove UnableToParseException when use this method with a pojo class has field which type is Instant

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -5,6 +5,7 @@ import java.lang.reflect.Field;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.text.NumberFormat;
+import java.time.Instant;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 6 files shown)


majst01:
Hi @flashmouse,

thanks for your contribution, would you please add
- a unit test covering this new functionality
- extend example in README.md
- add a CHANGELOG.md entry

other than that looks good.

flashmouse:
master branch version now is 2.16-SNAPSHOT and CHANGELOG latest change is 2.15 ,should I add 2.16 entry for CHANGELOG?

majst01:
yes of course

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -263,10 +264,11 @@ private void addFieldByAttribute(final Object pojo, final Field field, final Col
 
         TimeColumn tc = field.getAnnotation(TimeColumn.class);
         if (tc != null && Instant.class.isAssignableFrom(field.getType())) {
```
flashmouse: what should I do if some filed has TimeColumn annotation but filed type not Instant?

PR #598, opened 2019-05-27
fix npe when has null value tag field in pojo

flashmouse:
use addFieldsFromPOJO method build point will cause npe when some tag field hasn't set any value.

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -273,7 +273,9 @@ public class Point {
         }
 
         if (column.tag()) {
-          this.tags.put(fieldName, (String) fieldValue);
+          if (fieldValue != null && !((String) fieldValue).isEmpty()) {
+            this.tags.put(fieldName, (String) fieldValue);
+          }
         } else {
           this.fields.put(fieldName, fieldValue);
         }
--- a/src/test/java/org/influxdb/dto/PointTest.java
+++ b/src/test/java/org/influxdb/dto/PointTest.java
@@ -548,6 +548,33 @@ public class PointTest {
 		pojo.time = null;
 	}
 
+	@Test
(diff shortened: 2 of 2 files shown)


src/main/java/org/influxdb/dto/Point.java
```diff
@@ -273,7 +273,9 @@ private void addFieldByAttribute(final Object pojo, final Field field, final Col
         }
 
         if (column.tag()) {
-          this.tags.put(fieldName, (String) fieldValue);
+          if (fieldValue != null && !((String) fieldValue).isEmpty()) {
```
majst01: I understand the null check, but why not adding if empty, that might be on pupose ?

flashmouse: sry I dont know what you mean exactly,can you give some code or example...

majst01: the old behavior just added the value of the field, now it is only added if not empty ?? What is wrong with an empty value, this is a API breakage.

flashmouse: influx tag must be a string key and string value pair.if some field has column annotation :
- This filed not a String value,I think it's just incorrect usage
- This filed is a String value, but user havn't set or just give it "",i think we just need not to set this tag in Point.

if this field is an empty String,influxdb-java will also throw exception:
unable to parse 'example,tag1= value1=123i 1559035795995000000': missing tag value

flashmouse: tag.getValue().replace("", "\\ ");

seems could resolved insert empty tag value.

flashmouse: hi @majst01 tell me if you have any idea, my application will use in production soon, i don't want to maintain another branch

nmazzon: any news?

Issue #599, opened 2019-05-27
Query with different time format does not work if database is not specified.


fredo994:
If you do not specify database in query (even thou you have specified database on the client level) method InfluxDBl#query(Query, TimeUnit) will not return any result because it will not send database information.

Steps to reproduce:

```java
InfluxDB influxDB = InfluxDBFactory.connect(host + ":" + port);
influxDB.setDatabase(database);
// queryResult will be empty.
QueryResult queryResult = influxDB.query(new Query("..."),  TimeUnit.NANOSECONDS)
```

Quick fix:
```java
InfluxDB influxDB = InfluxDBFactory.connect(host + ":" + port);
influxDB.setDatabase(database);
// queryResult will be empty.
QueryResult queryResult = influxDB.query(new Query("...", "database"),  TimeUnit.NANOSECONDS)
```

Permanent fix:
Change into "InfluxDBImpl#query(Query, TimeUnit)" into:
```java
public QueryResult query(final Query query, final TimeUnit timeUnit) {
   String db = query.getDatabase();
    if (db == null) {
        db = this.database;
    }
    Call<QueryResult> call = null;
    if (query instanceof BoundParameterQuery) {
        BoundParameterQuery boundParameterQuery = (BoundParameterQuery) query;
        call = this.influxDBService.query(db,
                TimeUtil.toTimePrecision(timeUnit), query.getCommandWithUrlEncoded(),
                boundParameterQuery.getParameterJsonWithUrlEncoded());
    } else {
        call = this.influxDBService.query(db,
                TimeUtil.toTimePrecision(timeUnit), query.getCommandWithUrlEncoded());
    }
    return executeQuery(call);
  }
```
Here database information is taken from client level if not given in query object.

majst01:
Good catch, are you willing to provide a PR with a test proving correct behavior ? I am happy to review and merge.

PR #600, opened 2019-05-28
Fix issue 599: Query with different time format does not work if database is not specified.

fredo994:
Fix for issue #599 - to use database information present in the client if not set in query.

I also changed TimeUtils#toTimePrecision to be in compliance with documentation.

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -923,7 +923,10 @@ public class InfluxDBImpl implements InfluxDB {
 
   private String getDatabase(final Query query) {
   	String db = query.getDatabase();
-  	return db == null ? this.database : db;
+  	if (db == null) {
+  		return this.database;
+	}
+  	return db;
   }
 
   private interface ChunkProccesor {
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -608,14 +608,13 @@ public class InfluxDBImpl implements InfluxDB {
   @Override
   public void query(final Query query, final int chunkSize, final BiConsumer<Cancellable, QueryResult> onNext,
                     final Runnable onComplete, final Consumer<Throwable> onFailure) {
-
     Call<ResponseBody> call;
     if (query instanceof BoundParameterQuery) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -926,6 +921,11 @@ public void dropRetentionPolicy(final String rpName, final String database) {
     executeQuery(this.influxDBService.postQuery(Query.encode(queryBuilder.toString())));
   }
 
+  private String getDatabase(final Query query) {
+  	String db = query.getDatabase();
+  	return db == null ? this.database : db;
```
majst01: I dont like elvis, please consider without

src/main/java/org/influxdb/impl/TimeUtil.java
```diff
@@ -60,7 +60,7 @@ public static String toTimePrecision(final TimeUnit t) {
     case MICROSECONDS:
       return "u";
     case NANOSECONDS:
-      return "n";
+      return "ns";
```
majst01: unrelated change, if this is a problem, please raise another Issue/PR

src/main/java/org/influxdb/impl/InfluxDBImpl.java
fredo994: ok

src/main/java/org/influxdb/impl/TimeUtil.java
fredo994: This is not a problem (it works in both cases), but I've notice that the code is not consistent with documentation.

majst01:
Checkstyle complains:

```
[ERROR] /usr/src/mymaven/src/main/java/org/influxdb/impl/InfluxDBImpl.java:925:3: File contains tab characters (this is the first instance). [FileTabCharacter]
```

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -39,6 +40,7 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.function.Consumer;
+import org.junit.runner.notification.RunListener;
```
majst01: Unused ?

fredo994: Sorry don't know how that got in there.

majst01:
looks good so far, can you please add a entry in the README to point to this ?

PR #601, opened 2019-05-28
Fix #596 - null safe tag's value

### LINKED ISSUE #596: does all tags defined in POJO need assign a value?
I created a POJO to insert into influxdb:

  @Measurement(name = "mymeasurement")
  public class Pojo {

    @Column(name = "tag1", tag =true)
    private string tag1;

    @Column(name = "tag2", tag = true)
    private String tag2;

    @Column(name = "value")
    private Double value;

    public String getTag1() {
      return tag1;
    }

    public void setTag1(String tag) {
      this.tag1 = tag;
    }
	
	public String getTag2() {
      return tag2;
    }

    public void setTag2(String tag) {
      this.tag2 = tag;
    }

    public Double getValue() {
      return value;
    }

    public void setDoubleObject(Double v) {
      this.value = v;
    }
  }

when i create a object with out set all tags, it failed with null exception:

Pojo pojo = new Pojo();
pojo.SetTag1("tag1value");
pojo.setValue(1);

Long t = Instant.now().toEpochMilli();
Point point = Point.measurementByPOJO(pojo.getClass()).addFieldsFromPOJO(pojo).time(t,TimeUnit.MILLISECONDS).build();
           
influxDb.write(getDBName(), getRPName(), point);

java.lang.NullPointerException

         at org.influxdb.dto.Point.escapeKey(Point.java:471)

         at org.influxdb.dto.Point.concatenatedTags(Point.java:432)

         at org.influxdb.dto.Point.lineProtocol(Point.java:420)

my question is does all tag need to be assign a value?

nmazzon:
fix #596

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -439,10 +439,12 @@ public class Point {
 
   private void concatenatedTags(final StringBuilder sb) {
     for (Entry<String, String> tag : this.tags.entrySet()) {
-      sb.append(',');
-      escapeKey(sb, tag.getKey());
-      sb.append('=');
-      escapeKey(sb, tag.getValue());
+      if(tag.getValue() != null) {  
+        sb.append(',');
+        escapeKey(sb, tag.getKey());
+        sb.append('=');
+        escapeKey(sb, tag.getValue());
+      }
     }
     sb.append(' ');
   }


nmazzon:
duplicate of #598

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -439,10 +439,12 @@ public String lineProtocol(final TimeUnit precision) {
...
     for (Entry<String, String> tag : this.tags.entrySet()) {
-      sb.append(',');
-      escapeKey(sb, tag.getKey());
-      sb.append('=');
-      escapeKey(sb, tag.getValue());
+      if(tag.getValue() != null) {  
+        sb.append(',');
+        escapeKey(sb, tag.getKey());
+        sb.append('=');
+        escapeKey(sb, tag.getValue());
```
majst01: Why should a null value not be allowed, same discussion with #598

PR #603, opened 2019-06-05
Issue602

### LINKED ISSUE #602: Bad database query in callback returns nothing/success
This `influxDB.query(Query("SHOW SERRIES"))` (Note the double R) is working as intended and throwing an InfluxDBException with a good error message 
> error parsing query: found SERRIES, expected ...

Working with a callback 
```java
influxDB.query(new Query("SHOW SERRIES"), result -> {
        println(result) // This code is executed
    }, error -> {
        error.printStackTrace()
    })
```
the result is that `null`is printed to stdout. I think this is incorrect behaviour, since it should be possible to handle any error. I don't know whether it is usefull to call the error callback or call the success callback with a QueryResult containing an error and no result.

If you want, i can fix this and make a pullrequest, but the behaviour should be defined first

maxemann96:
This PR fixes #602


src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -19,25 +19,13 @@
...
-import java.util.Set;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
+import java.util.*;
```
majst01: can you please keep individual imports

src/test/java/org/influxdb/InfluxDBTest.java
```diff
@@ -175,24 +163,38 @@ public void testBoundParameterQuery() throws InterruptedException {
...
+   * Tests for callback query with a failure.
+   * see Issue #602
+   */
+  @Test
+  public void testCallbackQueryFailureHandling() {
+    final AsyncResult<QueryResult> res = new AsyncResult<>();
+
+    this.influxDB.query(new Query("SHOW SERRIES"), res.resultConsumer, res.errorConsumer);
+
+    Assertions.assertThrows(InfluxDBException.class, res::result,
```
majst01: is it also possible to test against expected Exception message ?

src/test/java/org/influxdb/InfluxDBTest.java
maxemann96: Fixed in 0254cbcec68ae936a42446fd2f5d01e7f38fd9ff

src/test/java/org/influxdb/InfluxDBTest.java
maxemann96: No problem, 0254cbcec68ae936a42446fd2f5d01e7f38fd9ff

majst01:
Checks failed, 3 out of 6, dunno why, please check

maxemann96:
Fixed

PR #616, opened 2019-08-21
Fix #614: drop NaN and infinity values from fields

### LINKED ISSUE #614: Client-side skipping of NaN values
A bit related to #537, but instead of validation I'd like to propose skipping NaNs to silence Telegraf/InfluxDB logs - they can't be reliably filtered, and some libraries (i.e. Kafka) use NaNs in their metric values.

Currently InfluxDB treats NaNs as invalid input, performs partial write and logs a warning. `InfluxDB#write` finishes with `InfluxDBException.UnableToParseException`.

I'd like to discuss having an option to silently drop invalid points during writes, or a clarification that clients shouldn't pass invalid values to points, so that logging in Apache Flink (apache/flink#8513) can be fixed in some way.

Crack:
Fox for #614 - drop NaN and +/- infinity values from fields while serializing Point to line protocol. They are unsupported by InfluxDB and all issues relating to them are old and got no progress.

I went with the simplest solution, assuming that if these values will be supported in the future then probably a change in driver will be required, unless InfluxDB aligns exactly with current Java serialization (NaN to U+FFFD - invalid character, and infinity to `∞`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixes
 
 - Add new annotation called TimeColumn for timestamp field in POJO bean, this can set Point time and precision field correctly, also avoid UnableToParseException when flush Point to influx.
+- Skip fields with NaN and infinity values when writing to InfluxDB
+  [Issue #614](https://github.com/influxdata/influxdb-java/issues/614)
 
 ## 2.15 [2019-02-22]
 
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -405,13 +405,15 @@ public class Point {
   }
 
   /**
-   * calculate the lineprotocol entry for a single Point.
+   * Calculate the lineprotocol entry for a single Point.
+   * <p>
+   * NaN and infinity values are silently dropped as they are unsupported:
+   * https://github.com/influxdata/influxdb/issues/4089
    *
-   * Documentation is WIP : https://github.com/influxdb/influxdb/pull/2997
+   * @see <a href="https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_reference/">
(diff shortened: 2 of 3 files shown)


majst01:
Hi @Crack,
thanks for the clean patch, would you please add a changelog entry as well ? thanks.

Crack:
Sure - added.

majst01:
Hi, 
one more nit, can you please clarify more (in the javadoc) where influxdb doc states that behavior

Crack:
Added. Officially InfluxDB supports IEEE-754 floating points, but documentation makes no mention on NaN and infinity being unsupported, so I linked to GitHub issue: influxdata/influxdb#4089

majst01:
But reading this issue, there is no response from some of the influx guys ? I am unsure if we should drop these values silently ?

Crack:
Use case I was thinking about when developing this is an application reporting Kafka metrics, namely: apache/flink#8513 (and some of my internal ones that use Spring Boot).

There are two goals I'd like to achieve:
1. Remove useless exception that is written to application log with each collection. That I can work around with selectively targeting log reporter and disabling it (but this can hide some real problems)
2. Remove useless error logs and make data usable in applications that expect valid line protocol (InfluxDB / Telegraf).

If we want some signaling of these values then I can do some changes around InfluxDB#write to throw an exception after a successful write to InfluxDB, and a flag to switch these off.

__________

Anyway, if InfluxDB starts supporting NaNs or infinities in the future, influxdb-java would need to be changed anyway, because current NaN serialization format outputs unicode garbage and that would need to be changed to something else.

majst01:
Thanks for the clarification and contribution !

Crack:
Thanks!

When do you plan to release a new version?

Crack:
@majst01 when do you plan to release a new version?

PR #625, opened 2019-10-01
Overriding Accept-Encoding header when gzip is disabled

muslinere:
- When gzip is disabled, do not add Accept-Encoding header to request. (Overrided with identity for Accept-Encoding)
 - In specific environments/deployments the addition of this encoding type causes issues


majst01:
Hi,

got your point, can you please add a test that ensures the correctness of the resulting header in this case, a additional entry in the CHANGELOG.md would also be nice.

majst01:
Any updates here ?

muslinere:
Sorry, I haven't had the time to figure out to implement a test for this. It wasn't as straight forward as I initially thought. If you have any ideas on how to test this it would be appreciated.

PR #628, opened 2019-10-23
fix(readme): add link to latest client lib

rhajek:


majst01:
Hi @rhajek,

i am happy to merge this, but i asked @ivankudibal multiple times to release the current state of influxdb-java with no reaction since month. 
So please release the actual master to maven-central as 2.6, after that i will merge this.

majst01:
I dont understand why this is required for a release, there is no dependency, but to bring this to an end i merge this.

bednar:
The client v2.16 is released...

majst01:
takes a while to be visible on maven central, right:
https://mvnrepository.com/artifact/org.influxdb/influxdb-java

Issue #643, opened 2019-12-21
InfluxDBFactory.connect() throws exception when using without username/password on okhttp 4.x.x


vinh0604:
One of my project is using `influxdb-java` to connect to an influxdb database without username and password.

After we upgraded `okhttp` to version 4.0.1, InfluxDB client starts failing with the following exception:
```
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method okhttp3.Credentials.basic, parameter username

	at okhttp3.Credentials.basic(Credentials.kt)
	at okhttp3.Credentials.basic$default(Credentials.kt:28)
	at okhttp3.Credentials.basic(Credentials.kt)
	at org.influxdb.impl.BasicAuthInterceptor.<init>(BasicAuthInterceptor.java:15)
	at org.influxdb.impl.InfluxDBImpl.<init>(InfluxDBImpl.java:151)
	at org.influxdb.impl.InfluxDBImpl.<init>(InfluxDBImpl.java:120)
	at org.influxdb.impl.InfluxDBImpl.<init>(InfluxDBImpl.java:184)
	at org.influxdb.InfluxDBFactory.connect(InfluxDBFactory.java:63)
```

It looks like from version 4, the method `Credentials.basic()` in `com.squareup.okhttp3:okhttp` does not accept null values for username and password anymore, but currently there is no way to not calling that method when initializing the InfluxDBImpl instance.

PR #644, opened 2019-12-21
Fix issue of connecting to influxdb without username/password on okhttp 4.x.x

vinh0604:
This pull request is to fix the bug mentioned in #643 by only add `BasicAuthInterceptor` to `OkHttpClient.Builder` when both username and password are not null.


majst01:
Hi @vinh0604 

thanks for this fix, i was experiencing the same when i tried to upgrade okhttp as part of my regular dependency updates. But i was to lazy to look into the root cause. So welcome !

actually all test are falling with:
```
[ERROR] /usr/src/mymaven/src/main/java/org/influxdb/impl/InfluxDBImpl.java:47: Using the '.*' form of import should be avoided - java.util.*. [AvoidStarImport]
```
this must be addressed first. I also have some small nits on the MR.

src/main/java/org/influxdb/impl/BatchProcessor.java
```diff
@@ -77,7 +77,7 @@ public Builder threadFactory(final ThreadFactory threadFactory) {
      *            is mandatory.
      */
     public Builder(final InfluxDB influxDB) {
-      this.influxDB = (InfluxDBImpl) influxDB;
+      this.influxDB = (InfluxDB) influxDB;
```
majst01: I think this cast can now be removed

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -44,10 +44,7 @@
 import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
```
majst01: this causes the build to fail, no wildcard imports please.

vinh0604:
Sorry, I didn't run stylecheck before pushing. Those issues were resolved now. 
Please check.

majst01:
Now all checks have passed, great. May i asked you to add a additional unit test case to ensure that the extra if statement is covered. I really appreciate your contribution !

vinh0604:
> Now all checks have passed, great. May i asked you to add a additional unit test case to ensure that the extra if statement is covered. I really appreciate your contribution !

Thanks, I added one more test case in `InfluxDBFactoryTest` as well as removing the if condition in the internal construct of `InfluxDBImpl` that is only used in test.

majst01:
Thanks for contributing.

PR #657, opened 2020-02-14
Allow to use BigInteger and BigDecimal for timestamps

### LINKED ISSUE #267: Nanosecond value in Point time field is stored in Long?
When creating points to write to influx using Java Instant:

```
        Instant stamp = Instant.now();
        Point point1 = Point.measurement("cpu")
                .time(stamp.getEpochSecond() * 1000000000 + stamp.getNano(), TimeUnit.NANOSECONDS)
                .addField("idle", 90L)
                .addField("system", 1L)
                .build();
```

From Java Instant documentation:
>The range of an instant requires the storage of a number larger than a long. To achieve this, the class stores a long representing epoch-seconds and an int representing nanosecond-of-second, which will always be between 0 and 999,999,999. 

So the previous code may overflow the Long field for time. Maybe time would be better as BigInteger?

J0s3f:
Allow to use BigInteger and BigDecimal for timestamps in the Point class to allow arbitrary timestamps with nanosecond precision.

Fixes #267

Won't break any existing code because it is still possible to use Long as timestamp.

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -4,6 +4,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
 import java.text.NumberFormat;
 import java.time.Instant;
 import java.util.Locale;
@@ -28,7 +29,7 @@ import org.influxdb.impl.Preconditions;
 public class Point {
   private String measurement;
   private Map<String, String> tags;
-  private Long time;
+  private Number time;
   private TimeUnit precision = TimeUnit.NANOSECONDS;
   private Map<String, Object> fields;
   private static final int MAX_FRACTION_DIGITS = 340;
@@ -89,9 +90,10 @@ public class Point {
    *
    */
   public static final class Builder {
+    private static final BigInteger NANOSECONDS_PER_SECOND = BigInteger.valueOf(1000000000L);
     private final String measurement;
     private final Map<String, String> tags = new TreeMap<>();
-    private Long time;
+    private Number time;
     private TimeUnit precision;
     private final Map<String, Object> fields = new TreeMap<>();
 
@@ -220,11 +222,12 @@ public class Point {
     /**
      * Add a time to this point.
      *
-     * @param timeToSet the time for this point
+     * @param timeToSet      the time for this point
      * @param precisionToSet the TimeUnit
      * @return the Builder instance.
      */
-    public Builder time(final long timeToSet, final TimeUnit precisionToSet) {
+    public Builder time(final Number timeToSet, final TimeUnit precisionToSet) {
+      Objects.requireNonNull(timeToSet, "timeToSet");
       Objects.requireNonNull(precisionToSet, "precisionToSet");
       this.time = timeToSet;
       this.precision = precisionToSet;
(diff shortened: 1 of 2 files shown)


src/main/java/org/influxdb/dto/Point.java
```diff
@@ -283,9 +286,17 @@ private void addFieldByAttribute(final Object pojo, final Field field, final Col
         TimeColumn tc = field.getAnnotation(TimeColumn.class);
         if (tc != null && Instant.class.isAssignableFrom(field.getType())) {
           Optional.ofNullable((Instant) fieldValue).ifPresent(instant -> {
-            TimeUnit timeUint = tc.timeUnit();
-            this.time = TimeUnit.MILLISECONDS.convert(instant.toEpochMilli(), timeUint);
-            this.precision = timeUint;
+            TimeUnit timeUnit = tc.timeUnit();
+            if (timeUnit == TimeUnit.NANOSECONDS || timeUnit == TimeUnit.MICROSECONDS) {
```
majst01: I would leave MILLISECONDS as the default path.

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -553,13 +564,37 @@ private void formatedTime(final StringBuilder sb, final TimeUnit precision) {
...
     }
-    if (precision == null) {
-      sb.append(" ").append(TimeUnit.NANOSECONDS.convert(this.time, this.precision));
-      return;
+    TimeUnit converterPrecision = precision;
+
+    if (converterPrecision == null) {
+      converterPrecision = TimeUnit.NANOSECONDS;
+    }
+    if (this.time instanceof BigInteger) {
```
majst01: I am a bit afraid of the performance impact, this is in the hot path and instanceof is not known for beeing super performant. Can we prove somehow, how big the performance impact might be ?

J0s3f: I created a benchmark to measure the impact. It can be found here: https://github.com/J0s3f/influx-point-benchmark
To run it on you own machine, simply download [the jar](https://github.com/J0s3f/influx-point-benchmark/releases/download/1.0/benchmarks.jar) and run it via `java -jar benchmarks.jar`

Here are the results on my machine: https://gist.github.com/J0s3f/6e6feb1cef2644bdb060576e99b1e91f

I'd say there is no meaningful impact of this change.

majst01: Its about 4% slower ! Given that users complain about performance already, i dont want to take extra without double check if this is not possible in another way

J0s3f: If you take the numbers from my benchmark, benchmarkNewPoint is the changed code so it is 4% faster (but with bigger error margin)

majst01: Cant image the extra code speeds up the processing ?

J0s3f: Benchmarks are just not accurate to that level on any PC that's running a full blown desktop OS.
I can try to increase the Forks in JMH but I don't have access to a bare metal machine I can use to run that for a few hours currently :(

And I changed one if in the existing code, this might have impact on the JIT compiler optimizations.

J0s3f: I made another benchmark using the settings `java -jar /benchmarks.jar -bm thrpt -f 5 -i 8 -si true -t max -wi 5` on a Quad-Core Intel Xeon VM that wasn't running anything else. (of course, I don't know what the host system was doing).

Here are the results: https://gist.github.com/J0s3f/379f9d2a1cfd6b77a5931b9422e80e42

majst01: OK, granted.

Can you please add some Unit tests to have these new logic covered as well: https://codecov.io/gh/influxdata/influxdb-java/compare/45f6cf44396acbf38161ab27b2c53ffbd2894fbf...fcc5a8fdc889d73af4a8495bad921db51c1dd33e/diff#D1-x558

J0s3f: Done

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -220,17 +222,30 @@ public Builder fields(final Map<String, Object> fieldsToAdd) {
...
 
+    /**
+     * Add a time to this point as Long.
+     * only kept for binary compatibility with previous releases.
+     *
+     * @param timeToSet      the time for this point as Long
+     * @param precisionToSet the TimeUnit
+     * @return the Builder instance.
+     */
+    public Builder time(final Long timeToSet, final TimeUnit precisionToSet) {
```
majst01: Why is this needed, Long is a Number right.

J0s3f: Yes but in Java the Argument type is part of the method signature. For people always recompiling their code, it is not needed. But if someone would compile with a version without this pull request and than later just switch the dependency without recompiling, the program will fail with a NoSuchMethodError

majst01: I'm getting slow in Java, since years i switch to go for many reasons.
Is this ready to be merged now ?

J0s3f: Yes, I think it is complete. The code coverage is strange, but this must be some error. I just added this one method and suddenly some entirely different classes have zero coverage.

PR #676, opened 2020-05-09
Fixes issue: InfluxDBResultMapper doesnt call POJO setter

### LINKED ISSUE #670: InfluxDBResultMapper doesnt call POJO setter
It is puzzling that `InfluxDBResultMapper` doesnt call POJO setter. 
My POJO looks like
```
class MyPOJO
{
    @Column(name = "time")
    private Instant time;

    //getter ignored
    public void setTime(Instant time) {  
        this.time = time.truncatedTo(ChronoUnit.SECONDS);
    }
}
```

Then I do something like
```
    final List<MyPOJO> resultEvents =  resultMapper.toPOJO(result, MyPOJO.class);
```
Assumption is `setTime` would be called for each point fetched. But it doesnt. It somehow bypasses it. `time` is private. Reflection in play here?
Expectation is that if I remove my setter, `toPOJO` call would fail. But it doesnt.

I have tried 2.17, 2.7 and 2.9 versions.

sjoshid:
Closes #670 

Things included in this PR:
* Call field setters instead of setting field values directly. 
* This PR maintains existing behaviour ie setting fields directly ignoring setters. 
* Priority is given to field setter if one exists. If not, field is set directly.

### DIFF
--- a/src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBResultMapper.java
@@ -21,6 +21,8 @@
 package org.influxdb.impl;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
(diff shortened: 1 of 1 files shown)


sjoshid:
@fmachado 
Im not sure if I need to run format sh. But when I do, I get a lot of diffs in most Java files. 
Anything I need to do with formattting?

fmachado:
@sjoshid thanks for submitting this PR.

> Im not sure if I need to run format sh.

I don't remember when was the last time that I had to run that shell script as I'm using the checkstyle.xml for a while.

But before getting into the formatting issues, could you please check why Travis build failed?

https://travis-ci.org/github/influxdata/influxdb-java/builds/685113586

sjoshid:
@fmachado 
Any feedback on this one?

sjoshid:
@majst01 
I have been waiting for @fmachado for feedback but havent received it. 
Can you provide one?

John9570:
whats the status on this? this would be a super useful feature to have

Issue #688, opened 2020-07-11
The client act in a synchronous manner when BatchProcessorr#queue is full.


simarpreets:
I went through the `BatchProcessor` implementation and noticed that when the batch processor is lagging behind and is not able to clear the `queue`,  **`InfluxDB.write(Point p)` becomes a blocking call even when one has configured it to write in an asynchronous manner.**
This is happening because the queue instance is of type LinkedBlockingQueue and `put(E e)` is used for enqueueing which blocks the thread trying to enqueue.
**IMO, this behaviour is very misleading, can block all the application threads who are trying to write whenever the write latencies on the influx-db server increases.**
I am using the `influxdb-java:2.15`
Below is the sample `Junit` test to simulate this.
```


import okhttp3.Interceptor;
import okhttp3.OkHttpClient;
import okhttp3.Response;
import org.influxdb.BatchOptions;
import org.influxdb.InfluxDB;
import org.influxdb.InfluxDBFactory;
import org.influxdb.dto.Point;
import org.influxdb.dto.Query;
import org.junit.Test;

import java.io.IOException;

public class InfluxTest {


    InfluxDB influxDB;

    @Test
    public void testConnection() {

        connect();

        for (int i=0;i<5000; ++i) {

            long before = System.currentTimeMillis();
            influxDB.write(Point.measurement("test").addField("a-field", 0).build());
            long after = System.currentTimeMillis();
            System.out.println(String.format("Time taken for writing %d times was %d", i, after-before));
        }

    }

    private void connect() {

            String influxDbConnectionURL = "http://localhost:8086";

            influxDB = InfluxDBFactory.connect(influxDbConnectionURL, new OkHttpClient.Builder().addInterceptor(new Interceptor() {
                @Override
                public Response intercept(Chain chain) throws IOException {
                    try {
                        Thread.sleep(10000);
                        System.out.println("interacted with influx");
                    } catch (InterruptedException e) {
                        throw new RuntimeException(e);
                    }
                    return chain.proceed(chain.request());
                }
            }));

            influxDB.query(new Query("CREATE DATABASE " + '"' + "test" + '"'));
            influxDB.setDatabase("test");

            influxDB.enableBatch(BatchOptions.DEFAULTS.exceptionHandler(
                (failedPoints, throwable) -> {
                })
            );

    }

}



```
The default batch size is 1000, so the 2000th write call will be blocked for ~10000ms.

majst01:
[...]

PR #689, opened 2020-07-16
Added an option to enable dropping of new points written after action queue exhaustion.

simarpreets:
Fix for #688

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 2.20 [unreleased]
 
+### Features
+- Add an option in `BatchOption` to prevent `InfluxDB#write` from blocking when actions queue is exhausted. [Issue #668](https://github.com/influxdata/influxdb-java/issues/688)
+
+
 ## 2.19 [2020-05-18]
 
 ## 2.18 [2020-04-17]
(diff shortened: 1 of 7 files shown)


majst01:
Build fails because of checkstyle errors, if these are fixed i can start reviewing

simarpreets:
@majst01 CI build is passing now, please review.

majst01:
LGTM, can you please add a Changelog entry and a section in the MANUAL.md with an useful example ?

simarpreets:
@majst01 Added a changelog entry and update MANUAL.md.
CI build failed because of some unrelated reason(seems like a network issue), will trigger it again after a while.

majst01:
@simarpreets  Thanks for your contribution !

simarpreets:
@majst01 when are you planning to release this?

sjoshid:
I think this is important change but can I nitpick?? What is the need for `dropActionsOnQueueExhaustion` boolean? I would have preferred to just have `droppedActionHandler`.

If `droppedActionHandler` is set, `dropActionsOnQueueExhaustion` is true. If null `dropActionsOnQueueExhaustion` is false. I hate having this separate flag that is meaningless.

Issue #684, opened 2020-06-24
the WriteApi does not check whether a tag cotains unauthorized characters ("\n","\r"...) 


AnesBendimerad:
InfluxDB has some restrictions on tag values : they cannot contain some characters such as:  line feed (\n), carriage return (\r). Currently, the Java client library does not check whether values of tags are safe (free of unauthorized characters) before writing them into InfluxDB. 
For example, I have a measurement classe called `Temperature`, and it contains a tag field `location`. When I put `location="mad\nrid"` and I write it using `WriteApi`, it will throw the following exception:  `partial write: unable to parse 'temperature,location=mad': missing fields dropped=0`. So a partial write will be done in the influxDB database: a new measurement `rid` will be mistakenly created, and it will contain only a part of the write. 
If a tag value contains '\r', the java client will not even throw an exception, and it will perform an incorrect write into the database.

I think it would be really useful to add a method that ensures that tag values are correct before sending them to InfluxDB through the line protocol, in order to avoid incorrect writes and anomalous behaviors.

majst01:
thanks for reporting, nice catch, if you would  create a merge request i will review and merge

BrentonPoke:
I think I could do this one once I find the appropriate classes. Has this been started by somebody already?

BrentonPoke:
> thanks for reporting, nice catch, if you would create a merge request i will review and merge

If you or anyone else could provide some feedback on my PR, that'd be great.

PR #691, opened 2020-08-01
#684: add tag checking for invalid characters

BrentonPoke:
This is to address a concern brought up in #684 where strings like "mad\nrid" with a \n character would bork a tag. I approached this through a set of two regular expressions:
`nameRegex = "([a-zA-Z0-9-_]+)"`
`valueRegex = "[\x00-\x7F]+"`

Names of tags I interpret as being more strict than values, which should probably be allowed the full range of ASCII characters. Current unknowns are weird preferences others might want to report in values. I personally think ASCII codes 0-127 should be sufficient for everybody's needs, but comment below.

These are applied through a utils subpackage created in the dto package. It's not intended to be used by the developer, but I wanted something slightly portable that can be used in dto classes in the future. <strike>I also added a new exception to the `InfluxDBException` wrapper so the modified tag methods in the `Point` and `BatchPoints` classes can through it to the developer to let them know what's wrong.</strike> There is a new exception, but to pass the unit tests, I've removed the throws for them.

I am new to this codebase, but am familiar with timeseries databases. Hope this is a good starting point.

### DIFF
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -15,7 +15,6 @@ import java.util.TreeMap;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import org.influxdb.BuilderException;
-import org.influxdb.InfluxDBException;
 import org.influxdb.annotation.Column;
 import org.influxdb.annotation.Measurement;
 import org.influxdb.annotation.TimeColumn;
@@ -136,11 +135,9 @@ public class Point {
      */
     public Builder tag(final Map<String, String> tagsToAdd) {
       for (Entry<String, String> tag : tagsToAdd.entrySet()) {
-        if(CheckTags.isTagNameLegal(tag.getKey())
-        && CheckTags.isTagValueLegal(tag.getValue()))
-        tag(tag.getKey(), tag.getValue());
-        else
-          continue;
+        if (CheckTags.isTagNameLegal(tag.getKey()) && CheckTags.isTagValueLegal(tag.getValue())) {
+          tag(tag.getKey(), tag.getValue());
+        }
       }
       return this;
     }
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -124,9 +124,6 @@ public class Point {
           && CheckTags.isTagValueLegal(value)) {
         tags.put(tagName, value);
       }
-      else {
-        throw InfluxDBException.buildExceptionForErrorState("tag name or value failed regex check");
-      }
       return this;
     }
 
@@ -139,7 +136,11 @@ public class Point {
      */
     public Builder tag(final Map<String, String> tagsToAdd) {
       for (Entry<String, String> tag : tagsToAdd.entrySet()) {
+        if(CheckTags.isTagNameLegal(tag.getKey())
+        && CheckTags.isTagValueLegal(tag.getValue()))
         tag(tag.getKey(), tag.getValue());
+        else
+          continue;
       }
       return this;
     }
--- a/src/main/java/org/influxdb/dto/Point.java
+++ b/src/main/java/org/influxdb/dto/Point.java
@@ -15,9 +15,11 @@ import java.util.TreeMap;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import org.influxdb.BuilderException;
+import org.influxdb.InfluxDBException;
 import org.influxdb.annotation.Column;
 import org.influxdb.annotation.Measurement;
 import org.influxdb.annotation.TimeColumn;
+import org.influxdb.dto.utils.CheckTags;
 import org.influxdb.impl.Preconditions;
 
 /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/influxdb/dto/utils/CheckTags.java
```diff
@@ -0,0 +1,45 @@
+package org.influxdb.dto.utils;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * {Purpose of This Type}.
```
majst01, line 7: Please describe the Type here.

src/main/java/org/influxdb/dto/utils/CheckTags.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * {Other Notes Relating to This Type (Optional)}
+ *
+ * @author Brenton Poke
+ *
+ */
+public final class CheckTags {
+  static final String NAMEREGEX = "([a-zA-Z0-9-_]+)";
```
majst01, line 15: Can you please point here to the official influxdb documentation where these restrictions are described and which influxdb versions are affected.

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -131,7 +135,9 @@ public Builder tag(final String tagName, final String value) {
      */
     public Builder tag(final Map<String, String> tagsToAdd) {
```
majst01, line 136: Why do we have different checks in these two tag methods ?

src/main/java/org/influxdb/dto/Point.java
```diff
@@ -116,7 +117,10 @@ private static void throwExceptionIfMissingAnnotation(final Class<?> clazz,
     public Builder tag(final String tagName, final String value) {
       Objects.requireNonNull(tagName, "tagName");
       Objects.requireNonNull(value, "value");
-      if (!tagName.isEmpty() && !value.isEmpty()) {
+      if (!tagName.isEmpty()
```
majst01, line 120: Missing test coverage for the case where tag or value is nil or empty

src/main/java/org/influxdb/dto/utils/CheckTags.java
BrentonPoke: There are no restrictions described in the documentation, which is the reason newline characters weren't being handled. I asked in #684 what other restrictions there should be, no feedback. These were reasonable starting points for that discussion.

src/main/java/org/influxdb/dto/Point.java
BrentonPoke: That was a mistake.

src/main/java/org/influxdb/dto/utils/CheckTags.java
majst01: OK, but i don't want to add restrictions here which are guessed, this will lead to a plethora of bug reports points not able to be written because of that.

BrentonPoke: Then I will only restrict newline and carriage return for #684, and maybe a new issue can be created if problems arise.

src/main/java/org/influxdb/dto/Point.java
BrentonPoke: This will be added.

BrentonPoke:
Sorry about the commit history. Never used Codecov before, but I learned a few things.

BrentonPoke:
Anybody available to review this one?

src/main/java/org/influxdb/dto/utils/CheckTags.java
BrentonPoke: This was added too.

BrentonPoke (commented):
All these changes have been made.

majst01:
TBH, 21 commits for a single change makes me nervous, and i am really unsure if i want to merge this PR.
@fmachado should at least check if this is worth merging.

BrentonPoke:
You can squash the commits. I tried to, but it didn't work as intended.

majst01:
I know that i can squash them, this is not my point.

BrentonPoke:
Then why else would what I've done not be worth merging?

majst01:
Thats why i asked @fmachado for additional review, its not useless.

fmachado:
@majst01 I'll take a look, thanks.

@BrentonPoke could you please share with us what InfluxDB reference documentation (or source code) you used to decide for that regexp? I couldn't find anything in your comments. :(

BrentonPoke:
> @BrentonPoke could you please share with us what InfluxDB reference documentation (or source code) you used to decide for that regexp? I couldn't find anything in your comments. :(

The first one just takes care of the issue brought up, which was that carriage return and newline are not properly ignored. The second one just makes sure that all characters are ascii values 0-127, which covers all of English and a number of symbols.
ASCII chart for values 0 through 127

fmachado:
> The first one just takes care of the issue brought up, which was that carriage return and newline are not properly ignored. The second one just makes sure that all characters are ascii values 0-127, which covers all of English and a number of symbols.

Apologies for not being clear with my first message: it's important for us to know that new code (1) is compliant with the line protocol (e.g. how InfluxDB handles what you name here "invalid characters") and (2) we are backward compatible with existing implementation (e.g. no unchecked exception, changes on public APIs and internal behavior like filtering non ASCII if they were accepted before).

We have to guide our implementation here based on how InfluxDB works. Could you please link here the **InfluxDB documentation** you used to support your implementation?

BrentonPoke:
The documentation [here](https://docs.influxdata.com/influxdb/v2.0/reference/syntax/line-protocol/) says no newline characters, but doesn't restrict ascii values for fields as far as I can tell. I can remove the value regular expression entirely, so is that ok?

fmachado:
@BrentonPoke yes, as it's written there:

> Line protocol does not support the newline character \n in tag values or field values.

Would be OK for you to submit a PR just with the relevant code and changes required for this fix?

BrentonPoke:
You mean throw away this commit history? I can change it on this PR right now. I guess I can transfer what's needed from this branch manually to a new one and make a new pull request, but I'm not sure what that solves that a squash merge couldn't.

BrentonPoke:
Since I already did it, I'll create the new pr from the other branch I have.

Issue #661, opened 2020-03-13
why are null/empty values mapped to 0 for a double ?


stoxalpha:
why are null/empty values mapped to 0 for a double ?

majst01:
can you please define more detailed what problem you are facing.

fdbatista:
I have same issue with a String attribute. I try to generate a Point with info from API, but if field is null, I get a partial write error when trying to write the point to Influx. Is there a way to set a defsupt value? Without manually using a setter, of course. Thanks.

fmachado:
@stoxalpha @fdbatista could you please add here:
- a test case or a piece of code that is failing;
- what you expected to happen;
- what happened.

fdbatista:
> @stoxalpha @fdbatista could you please add here:
> 
> * a test case or a piece of code that is failing;
> * what you expected to happen;
> * what happened.

Ok, here is my case:

I consume a remote API endpoint and convert this data into an InfluxDB Point via a custom POJO.

The API response sometimes comes with some null values, something like:
`{
  id: 1,
  title: "White Fang",
  author: "",
  isbn: "00234212"
}`

In this case, when I use the `Point.measurementByPOJO` function, the `author` field becomes `null` instead of empty string, so when trying to persist this point, InfluxDB engine says that it cannot parse this object, because the point gets generated like `title="White Fang",author=,isbn="00234212"`

Is it clear enough?

Thanks for your feedback.

PR #693, opened 2020-08-16
Request for annotation for null values on pojo fields of type String.

BrentonPoke:
A `@Default` annotation was requested to protect pojos coming back with null strings in #661. I interpreted it to mean something along the lines of what's in the unit test. Would appreciate feedback.

### DIFF
--- a/src/main/java/org/influxdb/annotation/Default.java
+++ b/src/main/java/org/influxdb/annotation/Default.java
@@ -9,4 +9,9 @@ import java.lang.annotation.Target;
 @Target(value = ElementType.FIELD)
 public @interface Default {
   String value() default "";
+  long longValue() default 0L;
+  int intValue() default 0;
+  double doubleValue() default 0.0d;
+  float floatValue() default 0.0f;
+  boolean boolValue() default false;
 }
--- a/src/main/java/org/influxdb/annotation/Default.java
+++ b/src/main/java/org/influxdb/annotation/Default.java
@@ -1,0 +1,12 @@
+package org.influxdb.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = ElementType.FIELD)
+public @interface Default {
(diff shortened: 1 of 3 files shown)


BrentonPoke:
Anyone with the ability to re-run jobs mind running the one that failed? All the others passed.

sranka:
@BrentonPoke I restarted the tests, they now pass. I am not an expert in this library, I took a quick look at the new code while restarting the tests. It seems to me
- the Default value always overrides any value of the field 
- the new test is not testing the Default annotation, it checks that the set values are transferred to point
- the new Default annotation requires sets a String value for other supported field types (Double, Integer, Float, BigDecimal, Instant, Boolean), which would certainly be unexpected 
- the relationship to #661 is not clear to me, it seems to me that it is already fixed by https://github.com/influxdata/influxdb-java/commit/6afc6ade28b073af2e53022316ada6a838b41815#diff-bd011207d3113c41cd8be4968d2321cb

BrentonPoke:
> @BrentonPoke I restarted the tests, they now pass. I am not an expert in this library, I took a quick look at the new code while restarting the tests. It seems to me that
> 
> * the Default value always overrides any value of the field
> * the new test is not testing the Default annotation, it checks that the set values are transferred to point
> * the new Default annotation always sets a String value even for other supported field types (Double, Integer, Float, BigDecimal, Instant, Boolean), which would certainly be unexpected
> * the relationship to #661 is not clear to me, it seems to me that it is already fixed by [6afc6ad#diff-bd011207d3113c41cd8be4968d2321cb](https://github.com/influxdata/influxdb-java/commit/6afc6ade28b073af2e53022316ada6a838b41815#diff-bd011207d3113c41cd8be4968d2321cb)

Thanks for responding, I'm taking another look at this since I originally did it early in the morning, but the comment [here](https://github.com/influxdata/influxdb-java/issues/661#issuecomment-599733357) is what was agreed upon as the desired change. I went off the end of this discussion that resolved the need-more-info label. For the test, I did struggle a bit in terms of trying to develop a test to check that the value is set to the default. I'm trying a Black Box testing approach, but this is my first time processing an annotation, so I'm doing this PR to learn how.

BrentonPoke:
Ok, I think it's ready for another look, if anyone wants.

PR #702, opened 2020-09-25
Cleaner version of #684

BrentonPoke:
This is to address a request from @fmachado for a new pull request with just the changes involving newline characters in tags. Carriage returns are also handled by the regular expression.


fmachado:
@BrentonPoke thanks again for your contribution!

When you create a PR, it's a good practice to limit your changes only to the issue you are fixing/improving.

Quoting a Gist that I found https://gist.github.com/mikepea/863f63d6e37281e329f8
> Try to only fix one issue or add one feature within the pull request. The larger it is, the more complex it is to review and the more likely it will be delayed. Remember that reviewing PRs is taking time from someone else's day.

Could you please review your PR and submit only the relevant code? Thank you!

BrentonPoke:
All the additional changes you're seeing are the result of the check style formatting script. It never would have made it through the Travis without it.

PR #703, opened 2020-09-27
Handling newline characters in tags

BrentonPoke:
Hopefully this is the last one I have to do to clean up #684. Another attempt was made at #702, but checkstyle made things...difficult.


BrentonPoke:
@fmachado So how is this one? Good enough?

PR #765, opened 2021-08-27
feat: `POST` query variants serializes `'q'` parameter into HTTP body

bednar:
## Proposed Changes

The `query` parameter for `post` variants of query is serialized into HTTP body.

It is usefull when you hit the `nginx` limit for `Request-URI` length, eq. HTTP error: `414 Request-URI Too Large`.

- https://docs.influxdata.com/influxdb/v1.3/guides/querying_data/#querying-data-using-the-http-api
- https://docs.influxdata.com/influxdb/cloud/reference/api/influxdb-1x/query/

Following query variants also respects the `requiresPost` settings:

1. `QueryResult query(Query, TimeUnit)`
1. `void query(Query, int, BiConsumer<Cancellable, QueryResult>,
                   Runnable, Consumer<Throwable>)`

## Checklist

- [x] CHANGELOG.md updated
- [x] Rebased/mergeable
- [x] A test has been added if appropriate
- [x] `compile-and-test.sh` completes successfully
- [x] Commit messages are in [semantic format](https://seesparkbox.com/foundry/semantic_commit_messages)
- [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/compile-and-test.sh
+++ b/compile-and-test.sh
@@ -20,6 +20,12 @@ docker run \
        --publish 8086:8086 \
        --publish 8089:8089/udp \
        --volume ${PWD}/influxdb.conf:/etc/influxdb/influxdb.conf \
+       --env DOCKER_INFLUXDB_INIT_MODE=setup \
+       --env DOCKER_INFLUXDB_INIT_USERNAME=my-user \
+       --env DOCKER_INFLUXDB_INIT_PASSWORD=my-password \
+       --env DOCKER_INFLUXDB_INIT_ORG=my-org \
+       --env DOCKER_INFLUXDB_INIT_BUCKET=my-bucket \
+       --env DOCKER_INFLUXDB_INIT_ADMIN_TOKEN=my-token \
        influxdb:${INFLUXDB_VERSION}-alpine
 
 echo "Starting Nginx"
@@ -38,6 +44,20 @@ docker run \
 echo "Running tests"
 PROXY_API_URL=http://nginx:8080/influx-api/
 PROXY_UDP_PORT=8080
+if [[ "$INFLUXDB_VERSION" == "2.0" ]]
+then
+    TEST_EXPRESSION="InfluxDB2Test"
+    # Wait to start InfluxDB
+    docker run --link influxdb:influxdb ubuntu:20.04 bash -c "apt-get update \
+      && apt-get install wget --yes \
+      && wget -S --spider --tries=20 --retry-connrefused --waitretry=5 http://influxdb:8086/ping"
+    # Create DBRP Mapping
+    BUCKET_ID=$(docker exec influxdb bash -c "influx bucket list -o my-org -n my-bucket | grep my-bucket | xargs | cut -d ' ' -f 0")
+    docker exec influxdb bash -c "influx v1 dbrp create -o my-org --db mydb --rp autogen --default --bucket-id ${BUCKET_ID}"
+    docker exec influxdb bash -c "influx v1 auth create -o my-org --username my-user --password my-password --read-bucket ${BUCKET_ID} --write-bucket ${BUCKET_ID}"
+else
+    TEST_EXPRESSION="*"
+fi
 
 docker run --rm \
        --volume ${PWD}:/usr/src/mymaven \
@@ -49,7 +69,7 @@ docker run --rm \
        --env INFLUXDB_IP=influxdb \
        --env PROXY_API_URL=${PROXY_API_URL} \
        --env PROXY_UDP_PORT=${PROXY_UDP_PORT} \
-       maven:${MAVEN_JAVA_VERSION} mvn clean install
+       maven:${MAVEN_JAVA_VERSION} mvn clean install -Dtest="${TEST_EXPRESSION}"
 
 docker kill influxdb || true
 docker kill nginx || true
--- a/src/main/java/org/influxdb/impl/InfluxDBImpl.java
+++ b/src/main/java/org/influxdb/impl/InfluxDBImpl.java
@@ -642,7 +642,11 @@ public class InfluxDBImpl implements InfluxDB {
       call = this.influxDBService.query(getDatabase(query), query.getCommandWithUrlEncoded(), chunkSize,
           boundParameterQuery.getParameterJsonWithUrlEncoded());
     } else {
-      call = this.influxDBService.query(getDatabase(query), query.getCommandWithUrlEncoded(), chunkSize);
+      if (query.requiresPost()) {
+        call = this.influxDBService.query(getDatabase(query), query.getCommandWithUrlEncoded(), chunkSize, null);
+      } else {
+        call = this.influxDBService.query(getDatabase(query), query.getCommandWithUrlEncoded(), chunkSize);
+      }
     }
 
     call.enqueue(new Callback<ResponseBody>() {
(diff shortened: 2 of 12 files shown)


bednar:
@majst01, InfluxDB 2 compatibility endpoints requires [Database and retention policy mapping](https://docs.influxdata.com/influxdb/v2.0/reference/api/influxdb-1x/dbrp/) for queries and writes. So there is no easy way how to use existing test cases with InfluxDB 2.

I've prepared separete test case for InfluxDB 2 - [InfluxDB2Test](https://github.com/influxdata/influxdb-java/blob/feat/query-as-body-for-influxdb-v2/src/test/java/org/influxdb/InfluxDB2Test.java)

You can start this testcase by: `INFLUXDB_VERSION=2.0 ./compile-and-test.sh`

compile-and-test.sh
```diff
@@ -38,6 +44,20 @@ docker run \
 echo "Running tests"
 PROXY_API_URL=http://nginx:8080/influx-api/
 PROXY_UDP_PORT=8080
+if [[ "$INFLUXDB_VERSION" == "2.0" ]]
+then
+    TEST_EXPRESSION="InfluxDB2Test"
+    # Wait to start InfluxDB
+    docker run --link influxdb:influxdb ubuntu bash -c "apt-get update \
```
majst01, line 51: missing version of the container image

src/test/java/org/influxdb/InfluxDB2Test.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @BeforeEach
+    public void setUp() throws NoSuchFieldException, IllegalAccessException {
+        String url = String.format("http://%s:%s", TestUtils.getInfluxIP(), TestUtils.getInfluxPORT(true));
+        influxDB = InfluxDBFactory
+                .connect(url, "my-user", "my-password")
+                .setDatabase("mydb")
+                .setRetentionPolicy("autogen");
+
+        // TODO - temporary solution before release - https://github.com/influxdata/influxdb/pull/22325
```
majst01, line 35: can we remove this as i see the PR is merged

majst01 (approved):

src/main/java/org/influxdb/impl/InfluxDBImpl.java
```diff
@@ -810,6 +835,18 @@ private boolean checkMessagePackSupport() {
...
+    Matcher matcher = APPLICATION_VERSION_PATTERN.matcher(version());
+    if (!matcher.find()) {
+      return false;
+    }
+    String[] versionNumbers = matcher.group().split("\\.");
+    final int major = Integer.parseInt(versionNumbers[0]);
+    final int minor = Integer.parseInt(versionNumbers[1]);
+    final int fromMinor = 3;
+    return (major >= 2) || ((major == 1) && (minor >= fromMinor));
+  }
```
bednar, line 848: The current version of InfluxDB 2 OSS doesn't return correct version. See:

```console
docker run --rm --publish 8086:8086 influxdb:2.0-alpine

curl -i -X GET http://localhost:8086/ping

> HTTP/1.1 204 No Content
> X-Influxdb-Build: cloud2
> X-Influxdb-Version: 
> Date: Tue, 31 Aug 2021 06:49:18 GMT
```

What do you think about changing the `checkFormEncodedQuerySupport` to optimistic variant? The encoding query into body will be supported for `unknown` version and all version greater or equal to `1.3`.

@rhajek, @majst01

src/test/java/org/influxdb/InfluxDB2Test.java
bednar: done

compile-and-test.sh
bednar: done

bednar:
@majst01, I had to change how is PR implemented:

- All query variants respects [`requiresPost`](https://github.com/influxdata/influxdb-java/blob/34dea3ae95e80304a3769443f98a774a953a7742/src/main/java/org/influxdb/dto/Query.java#L67) settings on `Query`
- POST variants serializes query into HTTP body 

Can you review this changes?

Thx

rhajek (approved):
LGTM

majst01 (changes requested):
So, does this mean that influxdb-java is now compatible with influxdb v2 ? If yes this must be reflected in the Readme, currently it says:

```
Note: This library is for use with InfluxDB 1.x. For connecting to InfluxDB 2.x instances, please use the influxdb-client-java client.
```

Please adjust accordingly

bednar:
@majst01 

> So, does this mean that influxdb-java is now compatible with influxdb v2 ? 

The `influxdb-java` is able to use [InfluxDB 1.x compatibility API](https://docs.influxdata.com/influxdb/v2.0/reference/api/influxdb-1x/), but this API supports only subset of v1 features, e.g. queries:

image

- https://docs.influxdata.com/influxdb/v2.0/query-data/influxql/#influxql-support

I think that current text is correct, because the client doesn't supports v2 API.

majst01:
> @majst01
> 
> > So, does this mean that influxdb-java is now compatible with influxdb v2 ?
> 
> The `influxdb-java` is able to use [InfluxDB 1.x compatibility API](https://docs.influxdata.com/influxdb/v2.0/reference/api/influxdb-1x/), but this API supports only subset of v1 features, e.g. queries:
> 
> image
> 
> * https://docs.influxdata.com/influxdb/v2.0/query-data/influxql/#influxql-support
> 
> I think that current text is correct, because the client doesn't supports v2 API.

Understood, but even that must be clarified in the Readme, please add a short description with a link to the official documentation.

bednar:
@majst01, note changed to `Note: This library is for use with InfluxDB 1.x and 2.x compatibility API. For full supports of InfluxDB 2.x features, please use the influxdb-client-java client.`

majst01 (approved):

majst01:
@bednar merged, and now all builds fail ?? Could you please check why.

majst01:
@bednar it still does not compile anymore, please fix this, otherwise i have to revert this

bednar:
@majst01 currently I'am fixing that issue, stay tuned.

The problem is caused by definition of GitHub PRs workflow: 

image

All 1.x builds are running agains 1.8...

bednar:
The PR with fix is prepared: #773

bentatham:
With this change, I am now seeing failures in 2.22 that I didn't see before in 2.21.
```
missing required parameter "q"
```
Using InfluxDB 1.8.10.
I don't think I'm doing anything too fancy.  Things like this:
```
BoundedParameterQuery.QueryBuilder
        .newQuery(queryString)
        .forDatabase(database)
        .bind("startTime", startTime)
        .bind("endTime", endTime)
        .create();
```
(which defaults to `requiresPost`=true)

Is this expected? What do I need to change?

bednar:
Hi @bentatham,

can you share a little bit more about your code? How do you initialise the client? How do you run your query? 

I just tried following code and everything works fine:

```java
try (InfluxDB client = InfluxDBFactory.connect("http://localhost:8086")) {

  BoundParameterQuery query = QueryBuilder
		  .newQuery("SELECT * FROM cpu WHERE time >= $startTime AND time <= $endTime")
		  .forDatabase("my_database")
		  .bind("startTime", "2015-08-18T00:00:00Z")
		  .bind("endTime", "2015-08-18T00:30:00Z")
		  .create();

  QueryResult result = client.query(query);

  System.out.println("result = " + result);
}
```

Regards

bentatham:
@bednar Thanks for getting back to me:

We are using influxDB behind vertx, but I don't think that should really matter unless something in the query object is massively not thread-safe and is consistently hitting a volatility issue or something really obscure.

Initialization:
```
influxDb = InfluxDBFactory.connect(
            configuration.getUrl().toString(), 
            configuration.getUser(), 
            configuration.getPassword(), 
            new OkHttpClient().newBuilder() 
              .connectTimeout(CONNECT_TIMEOUT_SECONDS_DEFAULT, TimeUnit.SECONDS) 
              .readTimeout(READ_TIMEOUT_SECONDS_DEFAULT, TimeUnit.SECONDS)
              .writeTimeout(WRITE_TIMEOUT_SECONDS_DEFAULT, TimeUnit.SECONDS) 
enableGzip();
```
Batching is not enabled.

And then invoked with:
```
influxDb.query(query, TimeUnit.NANOSECONDS)
```

Maybe something is different about the `query` that specifies units?  (My unit tests still fail with removing that, but that might be because of the units themselves, so I'll keep digging here too.)

bednar:
> @bednar Thanks for getting back to me:
> 
> We are using influxDB behind vertx, but I don't think that should really matter unless something in the query object is massively not thread-safe and is consistently hitting a volatility issue or something really obscure.

So the `vertx` performs as a proxy? Is there a support for query in the HTTP body?

> Initialization:
> 
> ```
> influxDb = InfluxDBFactory.connect(
>             configuration.getUrl().toString(), 
>             configuration.getUser(), 
>             configuration.getPassword(), 
>             new OkHttpClient().newBuilder() 
>               .connectTimeout(CONNECT_TIMEOUT_SECONDS_DEFAULT, TimeUnit.SECONDS) 
>               .readTimeout(READ_TIMEOUT_SECONDS_DEFAULT, TimeUnit.SECONDS)
>               .writeTimeout(WRITE_TIMEOUT_SECONDS_DEFAULT, TimeUnit.SECONDS) 
> enableGzip();
> ```
> 
> Batching is not enabled.
> 

I tried it with a direct connection to InfluxDB and everything works fine.

> And then invoked with:
> 
> ```
> influxDb.query(query, TimeUnit.NANOSECONDS)
> ```
> 
> Maybe something is different about the `query` that specifies units? (My unit tests still fail with removing that, but that might be because of the units themselves, so I'll keep digging here too.)

It doesn't matter.

mschwore:
I'm seeing the same behavior as @bentatham 

I submitted issue [!848](https://github.com/influxdata/influxdb-java/issues/848) describing the problem.

PR #769, opened 2021-09-12
improve error msg obscure when connect url port miss

camilesing:
see https://github.com/influxdata/influxdb-java/issues/768


majst01 (changes requested):
You should leave formatting as is, and change only what is required what you want to achieve. Otherwise reviewing is not possible.

Also "*" imports are not accepted in the project

camilesing:
Ok, I did  @majst01

src/main/java/org/influxdb/InfluxDBFactory.java
```diff
@@ -103,6 +109,25 @@ public static InfluxDB connect(final String url, final String username, final St
...
     Preconditions.checkNonEmptyString(username, "username");
     Objects.requireNonNull(client, "client");
+    checkUrl(url);
     return new InfluxDBImpl(url, username, password, client, responseFormat);
   }
+
+  /**
+   * Check url is legal
+   */
+  private static void checkUrl(final String url) {
```
majst01, line 119: method should also be part of Preconditions

camilesing: got it

majst01:
Will review after release in the next days, be patient

PR #773, opened 2021-09-13
fix: continuous integration

bednar:
1. Fixed sys env expansion in GitHub workflow
2. Fixed flaky tests:
     - `BatchOptionsTest.testJitterDuration`
     - `MessagePackBatchOptionsTest.testWriteWithRetryOnRecoverableError`
3. Fixed `MessagePackInfluxDBTest.testChunkingQueryPost` - MessagePack response doesn't contains last empty response


majst01:
I have seen you pushed directly to master, please never do this !

bednar:
@majst01 The GitHub Checks pass five times, so I think that the flaky tests are fixed.

majst01 (approved):

PR #774, opened 2021-09-15
jdk 17 is the next LTS

majst01:
merge after release


bednar (commented):
There is also jdk 16 in `mvn.sh`: https://github.com/influxdata/influxdb-java/blob/e2ed4ea587f961ba628f6e08694cd0f3b04ffe24/mvn.sh#L11

majst01:
> There is also jdk 16 in `mvn.sh`:
> 
> https://github.com/influxdata/influxdb-java/blob/e2ed4ea587f961ba628f6e08694cd0f3b04ffe24/mvn.sh#L11

done

bednar (approved):

PR #784, opened 2021-11-15
chore(ci): add InfluxDB 2.1 to pipeline

bednar:


src/test/java/org/influxdb/BatchOptionsTest.java
```diff
@@ -466,7 +466,7 @@ public void testHandlerOnRetryImpossible() throws InterruptedException {
 
       writeSomePoints(influxDB, 1);
 
-      Thread.sleep(200);
+      Thread.sleep(500);
```
majst01, line 469: Why this ?

bednar: Because this test is unstable and sometimes fail at next assertion: https://github.com/influxdata/influxdb-java/runs/4210575990?check_suite_focus=true

Maybe the better solution will be use [VerificationWithTimeout](https://site.mockito.org/javadoc/current/org/mockito/verification/VerificationWithTimeout.html):

```suggestion
      verify(mockHandler, timeout(500).times(1)).accept(any(), any());
```

majst01: looks promising

majst01 (approved):

src/test/java/org/influxdb/BatchOptionsTest.java
bednar: Fixed

PR #814, opened 2022-03-07
Revert "Bump checkstyle from 9.3 to 10.0"

majst01:
Reverts influxdata/influxdb-java#813 because jdk8 does not work anymore.

other idea would be to remove jdk8 from the CI matrix, but i think this is still a relevant jdk.


bednar (approved):
I agree that `jdk8` is still relevant.

PR #972, opened 2023-12-02
Default column name to pojo field name, and add `allFields` mode to `Measurement` annotation

### LINKED ISSUE #971: Default column name to pojo field name, and add `allFields` mode to `Measurement` annotation
It would be nice if `@Column` could be used without a name, in which case the field name would be used as the column name.

In addition, when a pojo has many fields, it would be nice to be able to annotate it with `@Measurement(name = "...", allFields = true)` instead of annotating every field with `@Column`.
Finally, an `@Exclude` annotation on fields would be helpful, for opting a field out in this mode.

I have implemented such support. Should I submit a PR?

eranl:
if `@Column` is used without a name, the annotated field's name is used as the column name.

In addition, when a pojo is annotated with `@Measurement(name = "...", allFields = true)`, then all its fields are loaded or saved.
Finally, an `@Exclude` annotation can be used for opting a field out in this mode.

Fixes #971.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -323,21 +323,40 @@ public class Cpu {
 }
 ```
 
-2. Add @Measurement,@TimeColumn and @Column annotations:
+2. Add @Measurement,@TimeColumn and @Column annotations (column names default to field names unless otherwise specified):
 
 ```Java
 @Measurement(name = "cpu")
 public class Cpu {
     @TimeColumn
-    @Column(name = "time")
+    @Column
     private Instant time;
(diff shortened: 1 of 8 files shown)


majst01:
Nice PR, thanks for you contribution, can you please also add a bit of documentation of you new features, best place would be [here](https://github.com/influxdata/influxdb-java/blob/master/MANUAL.md).

If this is done, i am happy to merge

majst01 (approved):

PR #979, opened 2023-12-05
Unify `Point`'s and `InfluxDBMapper`'s POJO saving logic

eranl:
Currently, there are two duplicate and somewhat divergent versions of this logic:
- `Point` doesn't check that values conform to field types and doesn't constrain allowed value types
- `InfluxDBMapper` uses a `time` field rather than `@TimeColumn`

This change unifies the logic and takes the stricter path, which means that it may break existing code that relies on the current lenient logic.


majst01 (approved):
Nice, thanks

PR #980, opened 2023-12-05
Support generic POJO super classes

eranl:
Also, simplify type checking.

Should I add this to the manual?

### DIFF
--- a/src/main/java/org/influxdb/impl/TypeMapper.java
+++ b/src/main/java/org/influxdb/impl/TypeMapper.java
@@ -49,10 +49,6 @@ public interface TypeMapper {
     return typeMapping::get;
   }
 
-  static TypeMapper of(TypeMapper innerMapper, TypeMapper outerMapper) {
-    return typeVariable -> map(typeVariable, innerMapper, outerMapper);
-  }
-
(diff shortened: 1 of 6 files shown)


majst01:
Support for Generics is for sure worth mentioning in the manual.

eranl:
> Support for Generics is for sure worth mentioning in the manual.

Done

src/main/java/org/influxdb/impl/TypeMapper.java
```diff
@@ -0,0 +1,83 @@
...
+        throw new IllegalStateException("Could not resolve type " + type);
+      }
+
+      return resolvedType;
+    }
+
+    return type;
+  }
+
+  static Type map(TypeVariable<?> typeVariable, TypeMapper innerMapper, TypeMapper outerMapper) {
```
majst01, line 73: This function is not covered in tests, is it even used ?

eranl: Sorry, this was a leftover. Deleted. Thanks for the catch!

majst01 (approved):

PR #1008, opened 2025-02-11
chore/update-ubuntu

tkyocum:
Bumping ubuntu runner to `-latest` as 20.04 is EOL and will be disabled 1 Apr 25.


majst01 (approved):
Thanks !

